PR #424, opened 2015-05-07
Convert to locking "with"/by context manager.

marksantcroos:
Per your (Andre's) suggestion.
Enjoy parsing the diff for real changes.

You can keep the branch alive after merge.


andre-merzky:
Did not do much of that parsing, that is hard to see with the indentation change... :P

PR #426, opened 2015-05-08
Refactor to re-create on disconnect.

marksantcroos:


andre-merzky:
Looks good to me.  I just realized though that one could probably get rid of `self.shells` altogether, and leave that functionality to the LM altogether...

marksantcroos:
Can you elaborate? I dont think you want to make the LM shell specific, right?

andre-merzky:
nono, not at all.  But it looks to me that the go adaptor keeps a pool/cache/... of shells around, one for each service endpoint (lets ignore that there is currently only one relevant service endpoint).  That is the same as having an LM pool for that endpoint of size 1, isn't it?  So instead of checking if the shell exists in `self.shells`, we can directly get a created or leased shell via `lm.lease()` ...  One does not even need to restrict the pool size: for serial operations, the pool will never be populated with more than one shell, for concurrent ops (multiple adaptor instances), a larger pool population comes automatically.  There is no need to replicate that on adaptor class level I think.

marksantcroos:
Ah, I think I see what you say.

I'll need to revisit this anyway, as there is still a race-condition between the alive check and the usage.
(Which is not only theoretical, I actually ran into it pretty quickly (assuming this is what happened))

andre-merzky:
Yes, of course, this only makes sense if it works :)  But then again, its good to find that error anyway...

PR #429, opened 2015-06-15
expose slurm nodelist option

luis-rr:
w, --nodelist=<node name list>

Request a specific list of node names. The list may be specified as
a comma-separated list of node names, or a range of node names (e.g.
mynode[1-5,7,...]). Duplicate node names in the list will be ignored.
The order of the node names in the list is not important; the node names
will be sorted by SLURM.


marksantcroos:
Hi Luis, thanks for working on this and creating a PR out of it!

However, your suggestion would deviate from the standard.
I suggest you take a look at GFD90 (http://ogf.org/documents/GFD.90.pdf), and in particular at the `CandidateHosts` entry.

I suspect that it would not change your diff too much if you rework it to make use of that field.

Please let us know if you have additional questions on the spec!

luis-rr:
Hi Mark,

Thanks for the heads up on CandidateHosts. I kind of expected to be violating the standard but don't know it well enough to find the correct attribute. I replaced the node_list option with CandidateHosts. Hope that's good enough.

Cheers,
Luis

luis-rr:
Hi Mark,
Could you let me know if you consider more changes are necessary or this should be done in a different way?
Thanks,
Luis

andre-merzky:
Hey Luis -- sorry for the long turnaround time.  I'm afraid Mark is offline 'til around the 12th of July.  Is the pull request urgent for you to get merged?  If so, I'll check it out -- otherwise I'd rather leave it to Mark, as he knows those parts of the SAGA code better than I do...

luis-rr:
Hi Andre,

No rush at all, I just wanted to check that you were not waiting on me for more changes. Let's wait for Mark.

Cheers,
Luis

src/saga/adaptors/slurm/slurm_job.py
```diff
@@ -473,6 +474,12 @@ def _job_run (self, jd) :
         if jd.attribute_exists("total_physical_memory"):
             job_memory = jd.total_physical_memory
 
+        if jd.attribute_exists("candidate_hosts"):
+            if isinstance(jd.candidate_hosts, (tuple, list)):
```
marksantcroos: Any particular reason why you also support a tuple? The spec says list and this is also what we generally support in SAGA.

marksantcroos:
Hi Luis,
Sorry for the delay, we had some deadlines and then I went on vacation.
I asked another question, and then the PR is good to go.
One request though, can you please change the target branch to devel though, as that is where it should end up.
Thanks!
Mark

luis-rr:
Hi Mark,

I'll close this PR and open a new one with devel as a target (unfortunately, github doesn't allow just changing an existing PR: https://github.com/isaacs/github/issues/18).

PR #435, opened 2015-07-20
General improvements to SLURM adaptor

marksantcroos:
Specifically to support Comet, also tested on Stampede.

### DIFF
--- a/src/saga/adaptors/slurm/slurm_job.py
+++ b/src/saga/adaptors/slurm/slurm_job.py
@@ -493,13 +493,12 @@ class SLURMJobService (saga.adaptors.cpi.job.Service) :
 
         # make sure we have something for number_of_processes
         if not number_of_processes:
-            self._logger.warning("number_of_processes not specified in submitted "
+            self._logger.debug("number_of_processes not specified in submitted "
                                  "SLURM job description -- defaulting to 1 per total_cpu_count! (%s)" % total_cpu_count)
-
             number_of_processes = total_cpu_count
 
         # make sure we aren't given more processes than CPUs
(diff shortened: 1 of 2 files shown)


andre-merzky:
Looks good to me, will try to test/merge today.  Does that require saga/rp releases to be in sync?  I assume so?

marksantcroos:
On stampede it should work without the SAGA changes I think. But for Comet it will need the SAGA changes indeed.

andre-merzky:
Ok, will test both.  Thanks.

src/saga/adaptors/slurm/slurm_job.py
```diff
@@ -516,8 +512,13 @@ def _job_run (self, jd) :
                               % (total_cpu_count, number_of_processes), 
                                  saga.NoSuccess, self._logger)
 
-        slurm_script += "#SBATCH --ntasks=%s\n"        % (number_of_processes)
-        slurm_script += "#SBATCH --cpus-per-task=%s\n" % (total_cpu_count/number_of_processes)
+        slurm_script += "#SBATCH --ntasks=%s\n" % (number_of_processes)
+
+        if total_cpu_count != number_of_processes:
+            slurm_script += "#SBATCH --cpus-per-task=%s\n" % (total_cpu_count / number_of_processes)
```
andre-merzky: Can this be a float?

marksantcroos: No, integer division only.

andre-merzky: ack

andre-merzky:
Thanks Mr. Rabbit!

PR #437, opened 2015-07-21
implement candidate_hosts for slurm adaptor

luis-rr:
list of host names which are to be considered
by the resource manager as candidate targets

taking advantage of the --nodelist sbatch option from SLURM:

--nodelist=<node name list>
Request a specific list of node names. The list may be specified as
a comma-separated list of node names, or a range of node names (e.g.
mynode[1-5,7,...]). Duplicate node names in the list will be ignored.
The order of the node names in the list is not important; the node names
will be sorted by SLURM.


luis-rr:
follow up to PR: https://github.com/radical-cybertools/saga-python/pull/429

marksantcroos:
Thanks Luis!

PR #442, opened 2015-07-29
Add PROCESSES_PER_HOST to all job adaptors.

marksantcroos:
Even if they don't use it directly, they should not stumble over it, which
currently breaks RP with many adaptors.

Maybe this needs to be extended with a warning or so if the adaptor has no
native support for this?

Andre, any reason not to merge this now and possibly create tickets to solve it properly?


andre-merzky:
Agree on the approach -- merge now, clean up eventually, create tickets for that.  Thanks!

PR #446, opened 2015-08-05
Add contexts to session at start of context list

jcohen02:
Recreation of PR #445 targetting devel branch rather than master:

When a session is created without the default=False property, it sets up contexts for all the SSH keys in a user's ~/.ssh directory. If many keys are present in that directory, this results in a number of registered security contexts. When attempting an SSH connection, these contexts are tried in order to see if a connection can be made. If many invalid keys are tried, the remote side may reject the login attempt due to too many incorrect authentication attempts being made.

When adding an explicit context using session.add_context(ctx), this is added to the end of the context list meaning that all other security contexts are tried first when attempting authentication and authentication will fail before reaching the explicitly added, valid context.

By adding contexts to the beginning of the context list, authentication attempts use the most recently added context first, allowing authentication to proceed successfully if the most recently added context is valid.

### DIFF
--- a/src/saga/session.py
+++ b/src/saga/session.py
@@ -48,8 +48,24 @@ class _ContextList (list) :
     #
     def append (self, ctx, session=None) :
 
+        ctx_clone = self._initialise_context(ctx, session)
+
+        # context initialized ok, add it to the list of known contexts
+        super (_ContextList, self).append (ctx_clone)
+        
+    def insert(self, index, ctx, session=None) :
+
+        ctx_clone = self._initialise_context(ctx, session)
+
+        # context initialized ok, add it to the list of known contexts
+        super (_ContextList, self).insert (0, ctx_clone)
+    
+    # Initialise a context to be added to the list of known contexts
+    # Returns a cloned, initialised context that can be added to the context
+    # list. 
+    def _initialise_context(self, ctx, session=None):
         if  not isinstance (ctx, saga.Context) :
-            raise TypeError, "appended item is not a saga.Context instance"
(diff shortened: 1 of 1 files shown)


marksantcroos:
Thanks for the trouble to recreate.

On the actual content, I see what you are trying to fix, and I would certainly like to address that issue.
However, there is probably a better way.

The implementation/semantics is that `Session.contexts` is a overloaded list of `Contexts` that does some additional checks, but other than that is a regular list.
`Session.add_context()` operates on that list by calling the `append()` method, which is clearly defined as _"add to the end of the list"_. Changing the semantics of that by pre-pending it on the list is not nice therefore.

As the SAGA spec only documents `add_context()`, my suggestion would be the rename the `append` method into `insert`, and have `add_context()` call `insert` instead of `append`.

Another (or additional) way would be the more _pythonic_ approach by having you directly operate on the `Session.contexts` list and thereby bypassing `add_context()`. (In that case, we probably still want to overload the `insert()` method just as we did with `append()`.

What do you think? Andre?

jcohen02:
No problem, I take your point re changing the semantics of `append` - given this,  `insert` seems like a more sensible option.

andre-merzky:
Yes, agree with that -- sounds cleaner...  Thanks!

marksantcroos:
@jcohen02 Would you mind doing that then? (If you commit to your branch it will automagically show up here)

jcohen02:
@marksantcroos That's fine, happy to do that.

marksantcroos:
Looks good, thanks!

PR #450, opened 2015-08-10
Passing ssh_timeout param to ssh ConnectTimeout option

### LINKED ISSUE #449: TypeError from make_dir() when specifying directory name that already exists
Missing tuple around format string parameters in `make_dir()` in [shell_file.py](https://github.com/radical-cybertools/saga-python/blob/407dae441f5031e09f0ea11c3d62466027933171/src/saga/adaptors/shell/shell_file.py#L757) results in a `TypeError` when trying to create a directory that already exists. Expected behaviour is for this to raise a `saga.NoSuccess` exception. e.g.

```
import saga
from saga.filesystem import Directory
ctx = saga.Context("ssh")
ctx.user_id = 'myuser'
ctx.user_key = '/home/myuser/.ssh/my_key'
sn = saga.Session()
sn.add_context(ctx)
dir = Directory('sftp://localhost/tmp/')
dir.make_dir('test')
dir.make_dir('test')
```

-> `TypeError: not enough arguments for format string`

andre-merzky:


andre-merzky:
Quick note that we'll delay this PR until early November -- sorry for the long wait... :/

andre-merzky:
This should be ready for merge, at long last.  Mark, can you please review?

PR #454, opened 2015-08-25
Feature/shell file staging

andre-merzky:
I'll open this  pull request, should have done a long time ago -- but this needs some testing love before it can be considered for merging...


src/saga/adaptors/shell/shell_job.py
```diff
@@ -1280,6 +1326,23 @@ def init_instance (self, job_info):
...
+    #
+    def _update_state (self, state) :
+
+        old_state = self._state
+
+        if  state == saga.job.DONE and \
+            old_state not in [saga.job.DONE, saga.job.FAILED, saga.job.CANCELED] :
+
+            # stage output data
+            # FIXME: _update_state blocks until data are staged.  That should not happen.
```
andre-merzky: This specifically needs some consideration before merge...

andre-merzky:
Mark, could you have a look at this one, too?

src/saga/adaptors/shell/shell_job.py
andre-merzky: Another comment on this, as clarification and update:

The problem here is that `job.get_state()` now can block until stage-out is done.  That is not ideal,  obviously -- but I think it is acceptable as intermediate step.

The underlying problem is that SAGA is, in its core, a synchronous API.  In other words, SAGA has no active component performing some actions for an application unless the application calls some SAGA API call or the other.  Thus, the output staging can only happen once we are in a SAGA API call which knows about the job state allowing for output staging,  which basically is `get_state()`.  To avoid blocking, the user can currently already call the async version of `get_state()`.

The 'real' solution (or 'a' real solution) would be to add some persistent background thread which manages job states, similar to what we are doing in RP.  This is not on the table right now, and I thus consider this PR (semantically) acceptable as an intermediate step.

marksantcroos:
This seems to work for absolute paths, thanks.
Related, how does one get a file inside the working directory? Does that need to be explicit or are there implicit ways?

andre-merzky:
At the moment I think that is only possible via the explicit route.  One could add support for relative paths (and thus pwd), by making sure that the sftp shell follows the `$PWD` of the ssh shell - but that is indeed not implemented.

marksantcroos:
Thanks

PR #456, opened 2015-08-25
Passing ssh_timeout param to ssh ConnectTimeout option

jcohen02:
Pass value set using the `SAGA_PTY_SSH_TIMEOUT` environment variable to the `ConnectTimeout` option of the SSH command when attempting to set up a connection to a remote node.

`SAGA_PTY_SSH_TIMEOUT` can be a floating point value but this is converted to an `int` prior to passing it to the SSH command.

This fix ensures that connect attempts don't hang for a long period when the remote node is not accessible.


andre-merzky:
I'll merge this now, thanks, and sorry for the delay!  Note that this will require https://github.com/radical-cybertools/radical.utils/pull/49 to function properly.

PR #463, opened 2015-09-14
disable -V for all crays

andre-merzky:


marksantcroos:
It requires a bit more as the test for "cray" is broken. Or you rather have a half fix now?

andre-merzky:
*sigh* - I wasn't aware of the test being broken,  thanks :/  But yes,  I think half a fix before we get others onto BW would be good...

marksantcroos:
As this is generally an improvement, I'll merge it anyway.

PR #465, opened 2015-09-18
Added passing of SSH options during PTYShell creation

jcohen02:
When creating a PTYShell object to undertake remote operations on a remote resource:

```conn = PTYShell('ssh://%s' % node_ip, session=self.session)```

it is sometimes useful to be able to pass SSH options to the underlying SSH command. 

For example, when SSHing to a cloud instance, it is possible that the IP address of the target resource is being re-used from another instance that has previously been connected to. In such cases, if the host key of the original node has been stored in the local known_hosts file, an "ssh key mismatch" error will occur and the connection will fail. This can be prevented by passing the "StrictHostKeyChecking=no" option to the SSH command line. A large number of other options are supported (see ssh manpage).

This patch enables support for passing SSH options when creating a PTYShell object. An arbitrary number of options may be passed as a dictionary of option:value pairs and all the specified options will be passed to the SSH command line as "-o" parameters. Options are passed as part of the standard options dictionary that may be passed to the ```PTYShell``` constructor, under the ```ssh_options``` key. For example:

```
opts = {}
opts['ssh_options'] = {'StrictHostKeyChecking':'no'}
conn = PTYShell('ssh://%s' % node_ip, session=self.session, opts=opts)
```

### DIFF
--- a/src/saga/adaptors/context/ssh.py
+++ b/src/saga/adaptors/context/ssh.py
@@ -211,18 +211,21 @@ class ContextSSH (saga.adaptors.cpi.context.Context) :
         # make sure we have can access the key
         api = self.get_api ()
 
-        key = None
-        pub = None
+        unexpanded_key = None
+        unexpanded_pub = None
         pwd = None
 
         
-        if         api.attribute_exists (saga.context.USER_KEY ) :
-            key  = api.get_attribute    (saga.context.USER_KEY )
-        if         api.attribute_exists (saga.context.USER_CERT) :
-            pub  = api.get_attribute    (saga.context.USER_CERT)
-        if         api.attribute_exists (saga.context.USER_PASS) :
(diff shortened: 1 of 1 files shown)


andre-merzky:
That looks good to me, thanks.  On the one hand, I am somewhat hesitant to make the pty layer more complex, OTOH it is really such a beats already that we should do what needs doing if that extents the life of pty, until there is a replacement in sight...

Let me take care of the conflicts and do some tests before merging...

andre-merzky:
Quick note that we'll delay this PR until early November -- sorry for the long wait... :/

andre-merzky:
Mark, would you please review the PBS-touching part of this one as well?  If that looks ok to you, I can take care of conflict resolving and merge.  Thanks!

marksantcroos:
Was this closed unintentionally?

PR #468, opened 2015-09-22
Fix/torque failures

### LINKED ISSUE #365: pbs job with exit_status != 0 doesn't mark job as failed

marksantcroos:
Fixes #365.

### DIFF
--- a/src/saga/adaptors/torque/torquejob.py
+++ b/src/saga/adaptors/torque/torquejob.py
@@ -731,46 +731,6 @@ class TORQUEJobService (saga.adaptors.cpi.job.Service):
 
     # ----------------------------------------------------------------
     #
-    def _retrieve_job(self, job_id):
-        """ see if we can get some info about a job that we don't
-            know anything about
-        """
-        # rm, pid = self._adaptor.parse_id(job_id)
-
-        # # run the PBS 'qstat' command to get some infos about our job
-        # if 'PBSPro_1' in self._commands['qstat']['version']:
-        #     qstat_flag = '-f'
-        # else:
-        #     qstat_flag ='-f1'
-        #
-        # ret, out, _ = self.shell.run_sync("unset GREP_OPTIONS; %s %s %s | "\
-        #         "grep -E -i '(job_state)|(exec_host)|(exit_status)|(ctime)|"\
-        #         "(start_time)|(comp_time)|(stime)|(qtime)|(mtime)'" \
-        #       % (self._commands['qstat']['path'], qstat_flag, pid))
-
-        # if ret != 0:
-        #     message = "Couldn't reconnect to job '%s': %s" % (job_id, out)
-        #     log_error_and_raise(message, saga.NoSuccess, self._logger)
-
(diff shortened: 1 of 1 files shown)


src/saga/adaptors/torque/torquejob.py
```diff
@@ -116,27 +116,21 @@ def log_error_and_raise(message, exception, logger):
...
+    elif torquejs == 'S': # "Job is suspended."
         return saga.job.PENDING
-    elif torquejs == 'W': # PBS Pro and TORQUE "Job is waiting for its execution time to be reached."
+    elif torquejs == 'W': # "Job is waiting for its execution time to be reached."
         return saga.job.PENDING
-    elif torquejs == 'R': # PBS Pro and TORQUE "Job is running."
+    elif torquejs == 'R': # "Job is running."
         return saga.job.RUNNING
-    elif torquejs == 'E': # PBS Pro and TORQUE "Job is exiting after having run"
+    elif torquejs == 'E': # "Job is exiting after having run"
```
andre-merzky: In Slurm, I opened a PR to have the equivalent of `E` (`CG`) interpreted as `CANCELED`, because at this point it is inevitable that the job will finish, and it can stay in `CG` rather long   for some reason, at least on BW and Stampede, and thus the RP shutdown is delayed quite a bit.  This is semantically not fully correct though.  Either way, we should keep it consistent I guess, and either purge the slurm PR, or also use that interpretation here and in the other adaptors.  

Thoughts?

src/saga/adaptors/torque/torquejob.py
```diff
@@ -867,17 +821,15 @@ def _job_get_info(self, job_id, reconnect):
                     val = val.strip()
 
                     # The ubiquitous job state
-                    if key in ['job_state']: # PBS Pro and TORQUE
-                        job_info['state'] = _torque_to_saga_jobstate(val)
+                    if key in ['job_state']:
+                        job_state = val
```
andre-merzky: This diff looks wrong, at least a little.  For one, that means that the fixed `_torque_to_saga_jobstate` it not used at all, and also, the line 873 then assigns the native state to the SAGA state.  Should the mapping move to there?

marksantcroos: You are right, good catch!
The mapping needs to go to line 825 indeed.

Alternatively I could move this all inside _torque_to_saga_jobstate by adding return_val to its parameters?

andre-merzky: Yeah, I guess that would also work, and would leave the interpretation localized...

marksantcroos: Ok, I did the semantic equivalent. That closes this thread of the PR.

src/saga/adaptors/torque/torquejob.py
andre-merzky: Let me convert this comment into a  ticket, that is not a priority right now.

andre-merzky:
Thanks!

PR #470, opened 2015-09-25
Recent NCSA suggestion introduced regression.

marksantcroos:
NCSA recently suggested us to move from cores to nodes. As BW uses a default ppn of 1, this lead to a RP slot count of 1 per node.


andre-merzky:
thanks for digging down into the abyss... :)

PR #473, opened 2015-09-26
move iterpos where it won't show up on iteration and duming of priv a…

andre-merzky:
…ttribs


marksantcroos:
Its gone from MongodB. Thanks!

PR #474, opened 2015-10-06
Feature/empty attrib init

andre-merzky:
add a `from_dict()` to the attrib interface (symmetric to the existing `to_dict()`).  This can be used to initialize attribute-inheriting objects via a dict.  Specifically, this is used in an RP branch to have pilot and units descriptions initialized by a dict.


marksantcroos:
Didn't break anything (yet :-)

PR #478, opened 2015-11-09
Fix #477, set session for shell job service

### LINKED ISSUE #477: Service does not propagate session object to shell adaptor
I am creating Service objects with *userpass* security context. While doing so I realized that folder creation fails on remote hosts. After investigation it seems that [the way](https://github.com/login?return_to=https%3A%2F%2Fgithub.com%2Fradical-cybertools%2Fsaga-python%2Fissues%2Fnew) *saga.base.Base* initializes the adaptor is the problem.

    self._init_task = self._adaptor.init_instance (adaptor_state, *args, **kwargs)

The above method call fails to propagate the *session* arguement to the underlying *shell adaptor* instance, as the *init_instance* [method](https://github.com/radical-cybertools/saga-python/blob/devel/src/saga/adaptors/shell/shell_file.py#L257) is implemented differently.

    def init_instance (self, adaptor_state, url, flags, session) :

To reproduce the problem take these steps:

    >>> js = saga.job.Service("ssh://joe2")
    >>> js.session == None
    True
    >>> ctx = saga.Context("ssh")
    >>> session = saga.Session()
    >>> session.add_context(ctx)
    >>> js = saga.job.Service("ssh://joe2", session=session)
    >>> js.session == None
    True

It seems that the session is always empty, no matter which context I do use. However, setting the session explicitly on the underlying adaptor solves the problem:

    >>> js._adaptor._set_session(session)
    >>> js.get_session()
    <saga.session.Session object at 0x286c950>

I can send you a PR in case you approve this issue.

mehdisadeghi:
This commit sets the session for `ShellJobService`. As far as I looked at the code, there are two methods called `_set_session` and `get_session` in `saga.adaptors.cpi.base.CPIBase` class which I think should be used in `init_instance` methods of `Service` classes to set the session on the service object correctly. However, this is not called in different job implementations such as pbs, sge and probably others. Shall we add this to all the other ones as well?

### DIFF
--- a/src/saga/adaptors/shell/shell_job.py
+++ b/src/saga/adaptors/shell/shell_job.py
@@ -455,6 +455,10 @@ class ShellJobService (saga.adaptors.cpi.job.Service) :
         self.jobs    = dict()
         self.njobs   = 0
 
+        # Use `_set_session` method of the base class to set the session object.
+        # `_set_session` and `get_session` methods are provided by `CPIBase`.
+        self._set_session(session)
+
         # if the rm URL specifies a path, we interprete that as shell to run.
         # Otherwise, we default to running /bin/sh (for fork) or the user's
         # login shell (for ssh etc).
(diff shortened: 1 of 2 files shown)


andre-merzky:
Hi Mehdi, 

yes, indeed, I think this should be uniformely used.  Thanks again for spotting this!

Would you mind recreating the pull request toward the devel branch, please (git/github do not allow  to change the PR target branch :( )?  Happy to merge it then!

Thanks, Andre.

mehdisadeghi:
Hi Andre,

I just created another PR toward devel branch, please have a look.

Issue #480, opened 2015-11-09
use _set_session() / get_session() in all adaptors


andre-merzky:
see #477/#479

mehdisadeghi:
Hey Andre,

Please see the above commit. I think if we add `session` as a property to the CPI base class, it will fix this problem for all the existing adapters. Having a property, we fill the corresponding `_session' in the base class with the same session which is desired. Moreover, since set/get session is a light weight operation, it makes sense to have it as a property and not a get/set method.

andre-merzky:
Hi Mehdi -- indeed, I think that should work nicely.  Neat! :D   Will you open a pull request toward devel?  I'd be happy to merge that in...

Thanks!

mehdisadeghi:
Good to hear that! Let's fix'em all :)

P.S. I just wanted to add one commit to the PR, but I don't know why there are three commits appearing there.

PR #481, opened 2015-11-09
Devel

mehdisadeghi:
This PR addresses incorrect sessions on `saga.filesystem.File` and `saga.filesystem.Directory` instances and is part of #480 . Without this fix, saga File and Directory instances will use the default session to carry on file operations, regardless of user provided session and security contexts (such as userpass).


andre-merzky:
:D

PR #484, opened 2015-11-24
Feature/external adaptors

jvegasbsc:
Hi

We are waiting for this feature to be available on the pypi version to release a version of Autosubmit  (https://pypi.python.org/pypi/autosubmit) based on SAGA. There is an estimation of when it will be available?

Thanks,
Javier Vegas

### DIFF
--- a/examples/jobs/localjob.py
+++ b/examples/jobs/localjob.py
@@ -23,7 +23,7 @@ def main():
 
         # Next, we describe the job we want to run. A complete set of job
         # description attributes can be found in the API documentation.
-        for i in range (100) :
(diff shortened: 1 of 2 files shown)


andre-merzky:
Hi @jvegasbsc,

thanks for the reminder!  Please give us another 2 days or so to test, merge and release -- but we should be able to get this out during the course of the week.

Best, Andre.

andre-merzky:
Thanks for merging, Mark.  I would not mind making a release at this stage -- do you see any problem with that?

marksantcroos:
Thats fine I guess.

PR #488, opened 2015-12-16
Feature/slurm reservation

andre-merzky:
IIRC, this was ready for merge back in Julich?


marksantcroos:
It's never tested/used. So if you want to merge it, then please give it a review.

andre-merzky:
Ah, ok -- I thought that was used in the tutorial, but IIRC, that reservation was changed into a specific queue at the end or something?  Anyway,  I don't have a slurm reservation anyway, and don't want to spend time on setting slurm up for testing right now, so I guess we'll leave this open 'til later then...

Thanks!

marksantcroos:
Given that I wrote it, I shouldnt be reviewing it :)
Anyway, lets wait on @iparask, now that we have a real use-case after all.

andre-merzky:
RIght - I did review it and it looks good, but could not test... :/  So yeah, lets wait for confirmation from Giannis...

andre-merzky:
Giannis, please open a ticket if that breaks...

PR #492, opened 2016-01-08
allow gsissh for condor job submission

andre-merzky:
This should only be merged after confirmation of functionality from RP devel (which at the moment stumbles over non-relative VE location specs)


marksantcroos:
Works fine, thanks.

Issue #494, opened 2016-01-11
support environment variables in session parameter file paths


jcohen02:
At present it seems that file paths passed to session parameters cannot include environment variables, e.g. the ability to specify a filename including $HOME.

For example, the following results in an error even though $HOME/.ssh/my_user_key is a valid key file:

```
import saga

ctx = saga.Context("ssh")
ctx.user_id='myuser'
ctx.user_key='${HOME}/.ssh/my_user_key'
s = saga.Session(default=False)
s.add_context(ctx)
```

This gives an exception similar to the following:

```
BadParameter: Cannot add context, initialization failed (ssh key inaccessible: ${HOME}/.ssh/my_user_key …
```

This would be a useful addition.

jcohen02:
Just to update the previous post, I note that in the Session class, the [documentation](https://github.com/radical-cybertools/saga-python/blob/5cfb4f23aa922ff8047ef005f29bc5bf4757621d/src/saga/session.py#L180-L195) suggests that $HOME can be used paths for certificate and key files so it looks like there may be a bug here?

It also looks like there are some errors in the [example](https://github.com/radical-cybertools/saga-python/blob/5cfb4f23aa922ff8047ef005f29bc5bf4757621d/src/saga/session.py#L180-L195) too - at least with the current version of the API - the creation of a Context appears to require a parameter - `c = saga.Context()` throws an error - and I assume the parentheses are missing from `s = saga.Session`. Also, should the file paths for `c.user_cert` and `c.user_key` be switched round?

andre-merzky:
Thanks for thePR, Jeremy!  I also fixed the docs as suggested, that will be merged soon I assume.

PR #496, opened 2016-01-12
Allow use of environment vars in ssh context key/cert file paths

jcohen02:
Suggested fix to #494 to address the problems described with this issue in handling file paths for SSH context key/cert file path properties that contain environment variables.


andre-merzky:
Thanks Jeremy,  that looks good!

jcohen02:
I noticed an issue with this fix if either the key or pub (cert) parameter is not specified. os.path.expandvars throws an exception if passed None. I've updated the fix and can issue a new PR with the further commit.

andre-merzky:
Yes,   sure, please do!  Thanks, Andre.

PR #499, opened 2016-01-13
Update to #496 to take account of empty key/cert parameter.

jcohen02:
Further update to PR #496, addressing issue #494, to resolve case where either the key or cert (pub) parameter are not specified.

### DIFF
--- a/src/saga/adaptors/context/ssh.py
+++ b/src/saga/adaptors/context/ssh.py
@@ -224,8 +224,14 @@ class ContextSSH (saga.adaptors.cpi.context.Context) :
             pwd  = api.get_attribute    (saga.context.USER_PASS)
 
         # Expand any environment variables in the key/pub paths
-        key = os.path.expandvars(unexpanded_key)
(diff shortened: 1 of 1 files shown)


andre-merzky:
Thanks!

Another option to fix it would be to change
```
unexpanded_key = api.get_attribute(saga.context.USER_KEY)
```

to 

```
unexpanded_key = api.get_attribute(saga.context.USER_KEY, '')
```

as the code really considers an empty string the same way as it considers 'None'.  I don't have a preference for either version, really, but wanted to mention it.  If you want to stick with this version, I'll merge it!

Best, Andre.

jcohen02:
I'm equally happy with either option, I guess your suggestions is somewhat nicer but I assigned None simply because this was what the value of key or pub would have been in the original implementation if no value had been provided.

I'm ok with merging as is but happy to make the change if you prefer.

andre-merzky:
Keeping it at `None` is probably safest,  who knows what side effect I am overlooking... :P

Thanks!

Issue #500, opened 2016-01-13
State callbacks not being made in PBS job adapter


jcohen02:
I'm experiencing issues when running jobs via the PBSJobService where a registered state callback function is not being called when expected.

I execute a job using the following approach:

```
jd = saga.job.Description()

### Set up job description parameters…

job = self.service.create_job(jd)
job.add_callback(saga.job.STATE, state_cb)

job.run()
job.wait()
```

I see one call of the callback function very soon after the job has been started where the state is set to `Pending` and the job ID value has been populated. After this I see no further calls to the callback function.

Debug output shows that the job monitoring thread is running. However, adding some additional debug output to the [job_state_monitor function](https://github.com/radical-cybertools/saga-python/blob/9f9b8edf8c0e488a3e6a87c95f52de4c5641cd11/src/saga/adaptors/pbs/pbsjob.py#L35-L102) in the [pbsjob.py](https://github.com/radical-cybertools/saga-python/blob/9f9b8edf8c0e488a3e6a87c95f52de4c5641cd11/src/saga/adaptors/pbs/pbsjob.py) module shows that the existing value for the job state `job_info['state']` is always the same as the new value `new_job_info['state']` which is obtained from the job service `new_job_info = self.js._job_get_info(job_id, reconnect=False)`.

Not sure if I'm doing something wrong in the way I'm configuring my jobs or setting up the callback but it looks like the job information in `jobs[job_id]` is being updated externally since it always contains the latest job status. As a test, removing the if statement `if  new_job_info['state'] != job_info['state']:` [here](https://github.com/radical-cybertools/saga-python/blob/9f9b8edf8c0e488a3e6a87c95f52de4c5641cd11/src/saga/adaptors/pbs/pbsjob.py#L78) causes the callback to be called on every iteration of the job monitor and changes in the state can be seen correctly.

jcohen02:
It looks like this issue is a result of [this](https://github.com/radical-cybertools/saga-python/blob/9f9b8edf8c0e488a3e6a87c95f52de4c5641cd11/src/saga/adaptors/pbs/pbsjob.py#L73-L80) code:

```
new_job_info = self.js._job_get_info(job_id, reconnect=False)
self.logger.info ("Job monitoring thread updating Job %s (state: %s)" \
                          % (job_id, new_job_info['state']))

# fire job state callback if 'state' has changed
if  new_job_info['state'] != job_info['state']:
    job_obj = job_info['obj']
[...]

PR #503, opened 2016-01-18
Fix for #500. Modified job state monitor to correctly identify state changes.

jcohen02:
Modified the PBS, PBSPro and Torque adaptors to resolve issue #500 that was preventing state callbacks from being correctly made when the job state monitor thread detects a change in state.

### DIFF
--- a/src/saga/adaptors/pbs/pbsjob.py
+++ b/src/saga/adaptors/pbs/pbsjob.py
@@ -69,18 +69,13 @@ class _job_state_monitor(threading.Thread):
                     # terminal state, so we can skip the ones that are 
                     # either done, failed or canceled
                     if  job_info['state'] not in [saga.job.DONE, saga.job.FAILED, saga.job.CANCELED] :
-                        
-                        # Store the current state since the current state 
-                        # variable is updated when _job_get_info is called
-                        pre_update_state = job_info['state']
 
                         new_job_info = self.js._job_get_info(job_id, reconnect=False)
-                        self.logger.info ("Job monitoring thread updating Job "
-                                          "%s (old state: %s, new state: %s)" % 
-                                          (job_id, pre_update_state, new_job_info['state']))
+                        self.logger.info ("Job monitoring thread updating Job %s (state: %s)" \
+                                       % (job_id, new_job_info['state']))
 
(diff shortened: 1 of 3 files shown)


marksantcroos:
Thanks Jeremy, we will go over it!

On the process, I should have made it more explicit. The `PBS` adaptor is read-only now, we only make changes to the `PBSPro` and `TORQUE`.

I actually have a branch that makes that more explicit that we should merge soon to prevent this from happening.
Would you be so kind to update your PR to undo the PBS changes this time?

jcohen02:
Apologies, I hadn't realised the `PBS` adaptor was now read-only. I've now reverted the changes to `pbsjob.py`.

I'll also switch to working with the PBSPro adaptor.

andre-merzky:
Many thanks, Jeremy!

marksantcroos:
Would this even extend to other adaptors?

jcohen02:
I'm not familiar with the other adaptors but a quick look suggests that the LSF adaptor is likely to be vulnerable to the same problem since it seems to have very similar job monitoring code to the PBS/PBSPro/Torque adaptors.

I had a quick look at the SGE, SLURM and Condor adaptors and these don't seem to use the same approach to job monitoring so I guess they're unlikely to suffer from this specific issue but I haven't had a chance to investigate in more detail.

I don't have an an environment to test the LSF adaptor but if you want, I can apply the same fix to the code there if someone else is able to review and test.

andre-merzky:
HI Jeremy -- sorry, I saw your comment only now, but I came to the same conclusion: only LSF seems to have a similar structure.  I opened #508 for a fix, it would be great if either you or Mark could review that.  I don't really know of a test env either though. Grace@yale and Yellowstone@NERSC use LSF, but I don't have access to either one...

jcohen02:
Hi Andre, no problem, I've taken a look at #508, it looks good to me. I wonder if this could be tested against an [OpenLava](http://www.openlava.org) deployment?

andre-merzky:
Hi Jeremy -- OpenLava looks interesting, thanks for the pointer!  I probably won't find the time though to deploy it for a test :/  So I guess we'll wait until a user comes along... :)

Cheers, Andre.

PR #505, opened 2016-01-18
Fixed string formatting error in shell_file adaptor

jcohen02:
A very minor one but I keep encountering this...


marksantcroos:
Thanks!

PR #507, opened 2016-01-19
Add comment leading to a new world order.

marksantcroos:


andre-merzky:
Who wouldn't want a new world order?

Issue #509, opened 2016-01-21
SSH context initialisation fails for encrypted key passwords containing spaces


jcohen02:
Not sure if this is a bug or just a shell-specific or platform-specific issue (Mac OS X) but it looks like ssh context initialisation fails when using an encrypted key that has a password containing spaces. This appears to be a result of the way the `subprocess.call` call is set up [here](https://github.com/jcohen02/saga-python/blob/9947658c1ec9568e073f432be8bff219d1012c57/src/saga/adaptors/context/ssh.py#L284)

To reproduce the error, try something like the following:

```
ssh-keygen -b 2048 -f mytestkey
```

When prompted for a password enter `my password`.  Now try and create an SSH context using this key:

```
import saga

s = saga.Session(default=False)
ctx = saga.Context("ssh")
ctx.user_id = 'testuser'
ctx.user_key = '/full/path/to/mytestkey'
ctx.user_pass = 'my password'

s.add_context(ctx)
```

This results in an error, the end of the stack trace showing:

```
/.../saga/session.pyc in _initialise_context(self, ctx, session)
     90             except se.SagaException as e :
     91                 msg = "Cannot add context, initialization failed (%s)"  %  str(e)
---> 92                 raise se.BadParameter (msg)
     93 
     94         return ctx_clone

BadParameter: Cannot add context, initialization failed (ssh key '/.../mytestkey' is encrypted, incorrect password (/.../saga/adaptors/context/ssh.py +294 (_initialize)  :  raise se.PermissionDenied ("ssh key '%s' is encrypted, incorrect password" % (key)))) (.../saga/session.py +92 (_initialise_context)  :  raise se.BadParameter (msg))
```

Prior to this, the help output for ssh-keygen appears which seems to suggest that it is being called with invalid parameters.

Recreating the command run by [this line](https://github.com/jcohen02/saga-python/blob/9947658c1ec9568e073f432be8bff219d1012c57/src/saga/adaptors/context/ssh.py#L284) on the terminal suggests that the password needs escaping:

```
sh -c 'ssh-keygen -y -f mytestkey -P my password'
```

Displays the `ssh-keygen` help.

```
sh -c 'ssh-keygen -y -f mytestkey -P my\ password'
```

Results in the command running correctly.

andre-merzky:
Indeed, you are right on target: escaping or quoting is missing here!  Thanks for spotting that and tracking it down!

vivek-bala:
PR #510 fixes this issue.

PR #510, opened 2016-01-21
Fix to ssh context to handle passwords containing spaces

jcohen02:
Proposed fix to #509. This takes spaces into account but not sure whether there are other characters that may also cause similar issues. Perhaps someone can review and see if this looks reasonable.

### DIFF
--- a/src/saga/adaptors/context/ssh.py
+++ b/src/saga/adaptors/context/ssh.py
@@ -5,7 +5,6 @@ __license__   = "MIT"
 
 
 import os
-import re
 
 import saga.context
 import saga.exceptions as se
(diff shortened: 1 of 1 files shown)


marksantcroos:
Would the following work also?

```
"ssh-keygen -y -f %s -P '%s' > /dev/null"
```

If so, then it will work for most other characters too.

andre-merzky:
I would agree with Mark here: quoting seems easier and more generic?

andre-merzky:
Hi Jeremy -- any comment?

jcohen02:
Hi, yes, I'd agree Mark's suggestion seems like a better option. I'll make the fix and update the branch.

andre-merzky:
Thanks!

jcohen02:
I've updated the fix as per Mark's suggestion and tested it. All seems ok.

andre-merzky:
Many thanks, Jeremy! :)

PR #514, opened 2016-02-04
Hotfix/staging tmp files

andre-merzky:
This fixes a FS level race when using pty level data staging in concurrent processes at high rates.  In the specific case, the user had two processes running which quickly submitted to slurm.  The slurm adaptor stumbled over missing batch  files now and then during staging.  The problem was ultimately caused by SAGA using the Python object IDs as part of a unique file name.  Those IDs however are not really unique, and clashed, so that one process removed tmp files belonging to the other.  This PR changes to use the tempfile module -- which is somewhat cumbersome semantically, but safe wrt the problem at hand.


marksantcroos:
+1 for using library calls.

PR #521, opened 2016-03-18
implement dir.exists

andre-merzky:


src/saga/adaptors/shell/shell_file.py
```diff
@@ -803,6 +803,31 @@ def get_size (self, tgt_in) :
...
+    #
+    @SYNC_CALL
+    def exists (self, tgt_in):
+
+        self._is_valid ()
+
+        cwdurl = saga.Url (self.url) # deep copy
+        tgt    = saga.Url (tgt_in)   # deep copy
+
+        ret, out, _ = self._command (" test -e '%s'" % tgt.path)
```
marksantcroos: How would this work for sftp and friends, they use the same code path, dont they?

andre-merzky: We always use an sh or ssh channel, and all commands go over that channel.  Only the actual movement of files is done via the sftp/scp layer.  The switch to the file transfer layer is done here: https://github.com/radical-cybertools/saga-python/blob/devel/src/saga/utils/pty_shell.py#L949

marksantcroos: Ah right, ok then!

PR #540, opened 2016-04-27
Feature/bw ppn xk

andre-merzky:
This needs discussing in the context of the RP PPN PR.


marksantcroos:
Yeah, I dont think I support this approach.

andre-merzky:
After discussion: this should go into CandidateHosts (see mem specs)

PR #544, opened 2016-04-27
Feature/osg optimizations

andre-merzky:
iteration on bulk and data staging support toward OSG.  This PR has not yet passed testing.


andre-merzky:
I would like to get this one merged for the upcoming release.  At the moment it contains three types of changes:

  - the obvious changes in the osg adaptor, which should not affect the other adaptors, and for which RP is the only known relevant user (?)
  - the job.Service now has a fallback implementation for run_job() (merged from `feature/osg_optimizations_run_job`)
  - smaller changes to examples, and removal of a redundant file (parser for transfer directives)

andre-merzky:
Hey Mark -- I would like to see this one getting ready for merge as well.  The branch has been in use by Matteo and me for quite a while now, and seems to hold up ok.

marksantcroos:
AFAIK its not compatible with RP devel.

andre-merzky:
Yeah, so the two-hop OSG staging is indeed still broken with devel, and I don't actually intent to fix this before switching to split_module.  FWIW, CU's without data staging work as expected, as do pilots then obviously.  We can discuss if merging still makes sense in this state, in terms of code consolidation, but that depends if anybody is using OSG, or expected to.  I am only aware of Matteo, and he is on the split route already...

andre-merzky:
Iain, I would like to defer this PR for one more release cycle: I'd rather do that merge in sync with RP's split module after all, as this SAGA branch has only seen exposure in that context.  I started a self review, but will need some time to do that somewhat thoroughly...

ibethune:
OK, makes sense.

PR #546, opened 2016-04-27
Feature/shared shell for files

andre-merzky:
This PR implements the use for shared channels for some additional methods, to reduce load on the ssh layer.


marksantcroos:
How should I interpret this? Were these operations all going to their own channel first?

andre-merzky:
Not each operation, but each object, which is a problem at scale.  This now uses the global connection pool.

PR #549, opened 2016-04-27
Fix/torque final differentation

andre-merzky:
7 months idle - what is supposed to happen to the branch?


andre-merzky:
Mark, are you using this branch?  Should that got into a test protocol, or can it be merged, in your opinion?

marksantcroos:
Doing this now would require more time than I have for this.

andre-merzky:
ok, np, we'll leave it open then.

PR #565, opened 2016-07-06
enable to have  in pbs jobs scripts by changing the echo command into…

wjlei1990:
… a verbose way

### DIFF
--- a/src/saga/adaptors/pbs/pbsjob.py
+++ b/src/saga/adaptors/pbs/pbsjob.py
@@ -316,15 +316,13 @@ def _pbscript_generator(url, logger, jd, ppn, gres, pbs_version, is_cray=False, 
     if gres:
         pbs_params += "#PBS -l gres=%s\n" % gres
 
-    # escape all double quotes and dollarsigns, otherwise 'echo |'
+    # escape all single quotes, otherwise 'echo |'
     # further down won't work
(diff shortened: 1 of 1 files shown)


src/saga/adaptors/pbs/pbsjob.py
```diff
@@ -690,7 +688,7 @@ def _job_run(self, job_obj):
         # (1) we create a temporary file with 'mktemp' and write the contents of 
         #     the generated PBS script into it
         # (2) we call 'qsub <tmpfile>' to submit the script to the queueing system
-        cmdline = """SCRIPTFILE=`mktemp -t SAGA-Python-PBSJobScript.XXXXXX` &&  echo "%s" > $SCRIPTFILE && %s $SCRIPTFILE && rm -f $SCRIPTFILE""" %  (script, self._commands['qsub']['path'])
+        cmdline = """SCRIPTFILE=`mktemp -t SAGA-Python-PBSJobScript.XXXXXX` &&  echo '%s' > $SCRIPTFILE && %s $SCRIPTFILE && rm -f $SCRIPTFILE""" %  (script, self._commands['qsub']['path'])
```
wjlei1990: If change the echo command to `echo '%s''`, then it will do in a verbose way without translating anything inside the string.

Issue #564, opened 2016-07-06
Improvements with method `_pbscript_generator()`


wjlei1990:
Hi,

I found there is improvements could be made in `adaptors/pbs/pbsjob.py`. In the function of `_pbsscript_generator()`, line https://github.com/radical-cybertools/saga-python/blob/devel/src/saga/adaptors/pbs/pbsjob.py#L319:

``` python
    # escape all double quotes and dollarsigns, otherwise 'echo |'
    # further down won't work
    # only escape '$' in args and exe. not in the params
    exec_n_args = workdir_directives + exec_n_args
    exec_n_args = exec_n_args.replace('$', '\\$')

    pbscript = "\n#!/bin/bash \n%s%s" % (pbs_params, exec_n_args)
    pbscript = pbscript.replace('"', '\\"')
```

The replacement of `$` is not good if you want to put some bash variable inside your pbs script...but you can't do that because `$` will be translated to `\$`.

The reason "$" needs to be replaced is because of line https://github.com/radical-cybertools/saga-python/blob/devel/src/saga/adaptors/pbs/pbsjob.py#L693:

``` python
        cmdline = """SCRIPTFILE=`mktemp -t SAGA-Python-PBSJobScript.XXXXXX` &&  echo "%s" > $SCRIPTFILE && %s $SCRIPTFILE && rm -f $SCRIPTFILE""" %  (script, self._commands['qsub']['path'])
        ret, out, _ = self.shell.run_sync(cmdline)
```

However, this issue could solved if change the `echo "%s" > $SCRIPTFILE` to `echo '%s' > $SCRIPTFILE`. Thus, the echo just dump the string content in a verbose way without interpolation. Accordingly, the escape rule should be changed to:

``` python
    # escape all single quotes, otherwise 'echo |'
    # further down won't work
    # only escape  in args and exe. not in the params
    exec_n_args = workdir_directives + exec_n_args

    pbscript = "\n#!/bin/bash \n%s%s" % (pbs_params, exec_n_args)
    exec_n_args = exec_n_args.replace('\'', '\\\'')
```

I have created a pull request: https://github.com/radical-cybertools/saga-python/pull/566

wjlei1990:
However, at runtime, it gives some warnings like:
```
2016-07-05 22:56:19,735: radical.saga.cpi    : MainProcess                     : MainThread     : WARNING : flush: [    5] [    1] (discard data: '$')
```
However, the program still runs OK.

Any clue?

[...]

PR #566, opened 2016-07-06
enable to have '$' in the pbs script by changing the echo command int…

wjlei1990:
…o a verbose way

### DIFF
--- a/src/saga/adaptors/pbs/pbsjob.py
+++ b/src/saga/adaptors/pbs/pbsjob.py
@@ -316,15 +316,13 @@ def _pbscript_generator(url, logger, jd, ppn, gres, pbs_version, is_cray=False, 
     if gres:
         pbs_params += "#PBS -l gres=%s\n" % gres
 
-    # escape all double quotes and dollarsigns, otherwise 'echo |'
+    # escape all single quotes, otherwise 'echo |'
     # further down won't work
-    # only escape '$' in args and exe. not in the params
     exec_n_args = workdir_directives + exec_n_args
-    exec_n_args = exec_n_args.replace('$', '\\$')
 
     pbscript = "\n#!/bin/bash \n%s%s" % (pbs_params, exec_n_args)
 
-    pbscript = pbscript.replace('"', '\\"')
+    pbscript = pbscript.replace('\'', '\\\'')
     return pbscript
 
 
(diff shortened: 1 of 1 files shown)


src/saga/adaptors/pbs/pbsjob.py
```diff
@@ -316,15 +316,13 @@ def _pbscript_generator(url, logger, jd, ppn, gres, pbs_version, is_cray=False,
     if gres:
         pbs_params += "#PBS -l gres=%s\n" % gres
 
-    # escape all double quotes and dollarsigns, otherwise 'echo |'
+    # escape all single quotes, otherwise 'echo |'
     # further down won't work
-    # only escape '$' in args and exe. not in the params
     exec_n_args = workdir_directives + exec_n_args
-    exec_n_args = exec_n_args.replace('$', '\\$')
```
wjlei1990: losing the '$' in bash script will be a big loss since all bash variable must have '$' to be used.

src/saga/adaptors/pbs/pbsjob.py
```diff
@@ -690,7 +688,7 @@ def _job_run(self, job_obj):
         # (1) we create a temporary file with 'mktemp' and write the contents of 
         #     the generated PBS script into it
         # (2) we call 'qsub <tmpfile>' to submit the script to the queueing system
-        cmdline = """SCRIPTFILE=`mktemp -t SAGA-Python-PBSJobScript.XXXXXX` &&  echo "%s" > $SCRIPTFILE && %s $SCRIPTFILE && rm -f $SCRIPTFILE""" %  (script, self._commands['qsub']['path'])
+        cmdline = """SCRIPTFILE=`mktemp -t SAGA-Python-PBSJobScript.XXXXXX` &&  echo '%s' > $SCRIPTFILE && %s $SCRIPTFILE && rm -f $SCRIPTFILE""" %  (script, self._commands['qsub']['path'])
```
wjlei1990: Change to `echo '%s'` will get copy things in a verbose way. Thus, we can include the '$' in the script.

andre-merzky:
This looks good to me -- Mark, any objection to merge?

marksantcroos:
Yes, there is still an unanswered question in the related issue.

andre-merzky:
Ah, I missed that.  For the record, the ticket is #564.

PR #572, opened 2016-07-15
Fixed run_job to run in service.py

mingtaiha:
Change function handle from run_job to run.

### DIFF
--- a/src/saga/adaptors/slurm/slurm_job.py
+++ b/src/saga/adaptors/slurm/slurm_job.py
@@ -598,13 +598,13 @@ class SLURMJobService (saga.adaptors.cpi.job.Service) :
                 message = "Couldn't create working directory - %s" % (out)
                 log_error_and_raise(message, saga.NoSuccess, self._logger)
 
-        
+
         # write script into a tmp file for staging
         self._logger.info ("SLURM script generated:\n%s" % slurm_script)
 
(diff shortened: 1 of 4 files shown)


andre-merzky:
Looking at the diff, the PR should be a one-line change, not a 12 line change ;)  You might want to use `git commit -p` to select specific patches for a commit.

More to the point though: does that actually work?  The adaptors are supposed to implement `job.Service.run_job()` -- `job.Job.run()` is a very different call.

mingtaiha:
Sorry for all those newlines. Idk if you wanted to review the code first or not since I've never touched the code. 

I can submit jobs fine. It also seemed like `run_job` was not implemented in the `JobService` class. I thought that the function call had to do with the `JobService` class of the appropriate resource manager, like `SLURMJobService` for SLURM. And both `run_job()` and `run()` belong to the `SLURMJobService` class, so I thought that there wouldn't be a problem. 

Upon a closer look, I do see that `self._adaptor.run()` is called in the `Job` class too. Maybe I was careless with the distinction between `job.Service.run_job()` and `job.Job.run()`?

mingtaiha:
I re-did the changes. Now, `run_job` 
1. First creates the job using `create_job(cmd)`, with `cmd` as a job description
2. Calls `run()` on the job
3. Returns the `job` object

PR #573, opened 2016-07-18
Removed duplicate get_name function in job class

jcohen02:
Removed duplicate `get_name` function in the `Job` class in `src/saga/job/job.py`.


andre-merzky:
thanks!

PR #576, opened 2016-07-24
Feature/session ctor

andre-merzky:
This should have been merged already, but has slipped off the radar -- apologies!


src/saga/__init__.py
```diff
@@ -15,7 +15,7 @@
 
 from   saga.task           import Task, Container
 from   saga.attributes     import Attributes, Callback
-from   saga.session        import Session
+from   saga.session        import Session, DefaultSession
```
marksantcroos: Why is DefaultSession exposed?

andre-merzky: Fair question!

The SAGA spec assumes that the session has no state, really, but just acts as a container for security credentials.  So there is no need to expose the default session, as its lifetime should have no side effects.

This is not exactly true in our implementation though: specifically, our connection caching is managed in the scope of a shell.  Jeremy stumbled over some problems with that, wrt. context lifetime management.  Exposing the default session is a step toward resolving those issues, as it gives direct access to the default session's context list (which was only accessible as a copy before).  Ultimately though, we also we may need a `session.close()` call to actively manage lifetime.

PR #579, opened 2016-07-30
Feature/versioned deps

andre-merzky:
This will require an additional step in our release procedure: if releasing the stack, we need to release buttom-up, and then adjust the dependent version numbers.  This probably speaks for coordinated version numbering across the stack...  Either way, it should resolve compatibility problems.

One could consider to make the version requirement more strict actually -- instead of using `>= x.y`, we could try to require `==x.y.*` .  That would make it possible to downgrade the whole stack with some confidence of workability...  Any opinion?


andre-merzky:
Possibly use range, and limit by major version change.

PR #587, opened 2016-09-29
Fix for issue #586, removing HOME env var from shell wrapper script file path

### LINKED ISSUE #586: Use of $HOME environment variable in shell adaptor prevents setup of wrapper.sh script
When using the shell adaptor on a resource that has not previously been used to run jobs, a wrapper.sh script is set up in `~/.saga/adaptors/shell_job/wrapper.sh`. The script is copied via scp/sftp to the remote node but the path used for the copy in [shell_job.py](https://github.com/radical-cybertools/saga-python/blob/e68f3706fbe40531793f8a4c97934e6d2a39af9c/src/saga/adaptors/shell/shell_job.py#L578) is of the form `$HOME/.saga/adaptors/shell_job/wrapper.sh`. When the copy is attempted, SSH interprets this as a relative path since it doesn't begin with a /. As a result, the target path used for the file copy is `/home/<username>/$HOME/.saga/adaptors/shell_job/wrapper.sh` and the copy fails with an error similar to the following:

```python
  File "/.../service_test.py", line 11, in <module>
    service = Service('ssh://localhost', session=session)
  File "/.../saga-python/src/saga/job/service.py", line 115, in __init__
    url, session, ttype=_ttype)
  File "/.../saga-python/src/saga/base.py", line 101, in __init__
    self._init_task = self._adaptor.init_instance (adaptor_state, *args, **kwargs)
  File "/.../saga-python/src/saga/adaptors/cpi/decorators.py", line 57, in wrap_function
    return sync_function (self, *args, **kwargs)
  File "/.../saga-python/src/saga/adaptors/shell/shell_job.py", line 514, in init_instance
    self.initialize ()
  File "/.../saga-python/src/saga/adaptors/shell/shell_job.py", line 580, in initialize
    self.shell.write_to_remote (src, tgt)
  File "/.../saga-python/src/saga/utils/pty_shell.py", line 835, in write_to_remote
    raise ptye.translate_exception (e)
saga.exceptions.DoesNotExist: file copy failed: Uploading /tmp/rs_pty_staging_lwJ9jW.tmp to /home/<user_id>/$HOME/.saga/adaptors/shell_job//wrapper.sh
remote open("/home/<user_id>/$HOME/.saga/adaptors/shell_job//wrapper.sh"): No such file or directory
sftp>  (/../saga-python/src/saga/utils/pty_shell.py +1015 (run_copy_to)  :  raise se.DoesNotExist._log (info['logger'], "file copy failed: %s" % str(out)))
```

This error can be reproduced using the following example code. Before running the code, remove any existing wrapper.sh script in `~/.saga/adaptors/shell_job/` to cause saga-python to try and recreate this file and trigger the problem:

```python
from saga import Context, Session
from saga.job import Service

ctx = Context("ssh")
ctx.user_id = '<MY_USERNAME>'
ctx.user_key = '<FULL PATH TO MY KEY FOR PASSWORDLESS LOCALHOST LOGIN>'
ctx.user_pass = ''
session = Session(default = False)
session.add_context(ctx)

service = Service('ssh://localhost', session=session)
```

jcohen02:
Fixes #586 by checking the start of the target file path when placing the shell job adapter `wrapper.sh` script into `~/.saga/...`. If the target path starts with `$HOME/` or `${HOME}/` then this is removed because sftp accepts a relative path as being relative to the authenticating user's home directory.

### DIFF
--- a/src/saga/adaptors/shell/shell_job.py
+++ b/src/saga/adaptors/shell/shell_job.py
@@ -575,6 +575,13 @@ class ShellJobService (saga.adaptors.cpi.job.Service) :
               # src = shell_wrapper._WRAPPER_SCRIPT % ({ 'PURGE_ON_START' : str(self._adaptor.purge_on_start) })
                 src = shell_wrapper._WRAPPER_SCRIPT
                 src = src.replace('%(PURGE_ON_START)s', str(self._adaptor.purge_on_start))
+                
+                # If the target directory begins with $HOME or ${HOME} then we
+                # need to remove this since scp won't expand the variable and
+                # the copy will end up attempting to copy the file to 
+                # /<path_to_home_dir>/$HOME/.....
+                if tgt.startswith("$HOME") or tgt.startswith("${HOME}"):
+                    tgt = tgt[tgt.find('/')+1:]
                 self.shell.write_to_remote (src, tgt)
 
         # ----------------------------------------------------------------------


andre-merzky:
Thanks Jeremy!

PR #593, opened 2016-11-10
OSG related condor fixes.

marksantcroos:
- Don't want our jobs to restart after eviction. 
- Fix requirements logic

Both got quite some airtime in my experiments.


andre-merzky:
Thanks Mark - I consider that sufficiently tested by your experiments.

PR #594, opened 2016-11-10
Feature/srm

marksantcroos:
This provides SAGA access to SRM storage.
Implemented as wrappers around the gfal2 command line tools, which are thus a requirement.
Also works over ssh and gsissh, so you only need access to a host that has these tools to use the SRM adaptor for 3rd party transfers.
(most) OSG nodes have the tools.


andre-merzky:
Thanks Mark!

PR #602, opened 2017-01-13
Feature/slurm bulk

andre-merzky:
This again enables the bulk calls for slurm, and makes the job name available after reconnect (and on the first pass, too).

This PR also cleans up some of the slurm code, which was, to be blunt, inefficient and sometimes plain wrong.  Some space for optimization remains, specifically when pulling for different job attributes, which individually results in remote queries, but I only added some comments for those.


src/saga/adaptors/slurm/slurm_job.py
```diff
@@ -522,24 +463,26 @@ def _job_run (self, jd) :
...
             slurm_script += "#SBATCH --error %s\n" % error
 
-        if  wall_time_limit:
+        if wall_time_limit:
             hours   = wall_time_limit / 60
             minutes = wall_time_limit % 60
             slurm_script += "#SBATCH --time %02d:%02d:00\n" % (hours, minutes)
 
-        if  queue:
+        print 'queue: %s' % queue
```
marksantcroos: You probably want to remove that.

andre-merzky: Done, thanks

marksantcroos:
Didn't test, but looks good. Thanks!

PR #603, opened 2017-01-13
Feature/torque bulk devel

andre-merzky:
This makes torque compatible with the RP split branches, by adding support for job names.


src/saga/adaptors/torque/torquejob.py
```diff
@@ -626,7 +625,8 @@ def _job_run(self, job_obj):
         """
 
         # get the job description
-        jd = job_obj.get_description()
+        jd       = job_obj.get_description()
```
marksantcroos: Not going to miss this :-)

marksantcroos:
Thanks!

andre-merzky:
:)  thanks!

PR #608, opened 2017-01-22
Feature/futures devel

andre-merzky:
This PR follows the API changes in RU, as documented in https://github.com/radical-cybertools/radical.utils/pull/95.


ibethune:
I think this is out-of-scope for 0.45 (correct me if wrong?)

andre-merzky:
Yes, indeed - this should be merged in sync with RU's `feature/managed_process` branch, which in turn should be in sync with RP's split branch.

andre-merzky:
This was closed because the diff is basically empty at this point.

PR #610, opened 2017-02-05
Cobalt adaptor implementation

itomaldonado:
First iteration of the Cobalt adaptor.


andre-merzky:
Many thanks, Manuel! :)

Issue #611, opened 2017-02-06
Slurm:  Getting info on an external job


markperri:
Is it possible to get info on a job that was run externally?  If I want to list all jobs I have running (that were started in another program, or if this program crashed and I'm restarting it).  I create a jobservice (jscomet) and use:

ids = jsComet.list()
        for job_id in ids:
             job = jsComet.get_job(job_id)

It throws an error because there's no prev_info for this external job:

  File "/home/mark/saga-python/src/saga/adaptors/slurm/slurm_job.py", line 893, in _job_get_info
    if prev_info.get('gone', False):
AttributeError: 'NoneType' object has no attribute 'get'

I added two checks to see if prev_info exists to get around this and it seems to work fine. I can get the job state without errors.  Is there a problem with using it this way?  Should I do it another way?

Starting at line 893 in src/saga/adaptors/slurm/slurm_job.py

>        if prev_info:
            if prev_info.get('gone', False):
                self._logger.debug("Job is gone.")
                return prev_info

        # curr. info will contain the new job info collect. it starts off
        # as a copy of prev_info (don't use deepcopy because there is an API
        # object in the dict -> recursion)
        curr_info = dict()
>        if prev_info:
            curr_info['job_id'     ] = prev_info.get('job_id'     )
            curr_info['job_name'   ] = prev_info.get('job_name'   )
            curr_info['state'      ] = prev_info.get('state'      )
            curr_info['create_time'] = prev_info.get('create_time')
            curr_info['start_time' ] = prev_info.get('start_time' )
            curr_info['end_time'   ] = prev_info.get('end_time'   )
            curr_info['comp_time'  ] = prev_info.get('comp_time'  )
            curr_info['exec_hosts' ] = prev_info.get('exec_hosts' )
            curr_info['gone'       ] = prev_info.get('gone'       )

Thanks,
Mark

marksantcroos:
In principle you should be able to reconnect to existing jobs, if that is your question.
You might have stumbled upon a bug then.

As a heads-up, our main developer is offline currently, so a more detailed reply will likely have to wait until next week.

andre-merzky:
Yeah, that looks like a bug, thanks for the report!  Feel free to submit a pull request with the workaround you seem to have in place, we would appreciate that!

andre-merzky:
PS.: I just saw that you joined github only recently.  Let us know if you need help creating a pull request, or if you would rather have us committing the fix! :)

markperri:
[...]

PR #612, opened 2017-02-07
check that prev_info exists before populating curr_info with its info…

markperri:
Added two checks for the existence of prev_info.  Previously if you tried to reconnect to a job by calling get_job(job_id) the Slurm adapter tried to pre-populated it with the results of the previous call.  I'm not sure if it should get some default values though, rather than just dict().


andre-merzky:
Thanks for the PR, this works as expected, and fixes #611

PR #624, opened 2017-03-09
Properly support SGE job name.

marksantcroos:
0.46 stack broke RP on SGE because of incorrectly implemented job name.


andre-merzky:
Yes, indeed - the stack now needs job name and bulk submission / state / cancel calls.  Thanks!

andre-merzky:
PS.: I looked over the commit already, just in case you wondered about the quick merge ;)

PR #625, opened 2017-03-10
add CANDIDATE_HOSTS for LoadLeveler

andre-merzky:


marksantcroos:
Euhm ... isn't this missing something?

andre-merzky:
Yes, a proper fix :)  This basically says that CH is supported, but does not actually support it.  It will stop a check from kicking in on an empty CH set.  The proper fix would be to let the check accept an empty attribute as not set. I opened #626 for this now.

marksantcroos:
Ah ok

PR #627, opened 2017-03-12
try to make RTD happy

andre-merzky:
see https://readthedocs.org/projects/saga-python/builds/5131031/


marksantcroos:
Shouldn't you include `sphinx.ext.imgmath` then?

andre-merzky:
Well, part of the message is:  `sphinx.ext.mathjax: other math package is already loaded`, that's why I just removed the pngmath package...  Sphinx seems to build ok locally with this change, we'll see what RTD thinks once it's merged...

marksantcroos:
Ok, thanks!

PR #630, opened 2017-04-19
Space is big. Really big. You just won't believe how vastly, hugely, mind-bogglingly big it is.

andre-merzky:
I mean, you may think it's a long way down the road to the chemist, but that's just peanuts to space.

-- Douglas Adams, The Hitchhiker's Guide to the Galaxy

Slurm at least is not able to grasp the emptiness of space...  This fixes the RP ticket #1319


mturilli:
Successfully tested on Stampede and reviewed.

AA919:
I knew that one day we would have broken Andre......

Issue #644, opened 2017-09-30
Slurm: -C, --constraint=list


breuera:
Similar to Bridges,
https://github.com/radical-cybertools/saga-python/blob/74fc38113477f22d1532aea120dc380fb677934d/src/saga/adaptors/slurm/slurm_job.py#L472
Cori requires setting `-C` in the job-script for the architecture:
http://www.nersc.gov/users/computational-systems/cori/running-jobs/batch-jobs/

Hard-coded this could be:

```
 468         if 'bridges' in self.rm.host.lower():
 469             # bridges requires '-C EGRESS' to enable outbound network
 470             # connections.
 471             # FIXME: this should be moved into a resource config file
 472             slurm_script += "#SBATCH -C EGRESS\n"
 473 
 474         if 'cori' in self.rm.host.lower():
 475           slurm_script += "#SBATCH -C knl\n"
```

andre-merzky:
thanks @breuera  - as ugly as the hostname specifica are, for now that's the easiest way to handle them.

Would you mind opening a pull request? Otherwise I'll add this in.

breuera:
Sure, see PR.

PR #645, opened 2017-10-01
CPU architecture on NERSC's Cori

breuera:
Solves #644.


andre-merzky:
Mapping to `CPU_ARCHITECTURE` makes sense.  Thanks for the PR!

PR #646, opened 2017-10-03
Fixed processes per host (#ranks/node) in SLURM-settings.

breuera:
Old:
```
if total_cpu_count and number_of_processes:
```
is always true (set before). 

```
int(total_cpu_count / number_of_processes)
```
gives `0` if `total_cpu_count` is not set manually (Slurm error).

----

New:

Only sets `--cpus-per-task` if `processes_per_host` is not set.
Having both, `--cpus-per-task` and `--ntasks-per-node` doesn't make much sense anyways.
Also `--cpus-per-task` is quite annoying since it requires setting it to the number of hyperthreads for a full node.


andre-merzky:
Indeed, that makes sense.  Thank you!

PR #652, opened 2018-01-17
Add gres SBATCH option

RajdeepKonwar:
Adds gres SBATCH option in order to support Generic Resources Scheduling (GRES) for gpu architectures.

### DIFF
--- a/docs/source/library/job/index.rst
+++ b/docs/source/library/job/index.rst
@@ -288,7 +288,7 @@ SAGA defines the following constants as valid job description attributes:
 .. autodata:: OPERATING_SYSTEM_TYPE
 .. data:: operating_system_type 
 
-   (Property) Same as attribute :data:`~saga.job.OPERATIN_SYSTEM_TYPE`.
+   (Property) Same as attribute :data:`~saga.job.OPERATING_SYSTEM_TYPE`.
 
 .. autodata:: CANDIDATE_HOSTS
 .. data:: candidate_hosts 
@@ -299,6 +299,11 @@ SAGA defines the following constants as valid job description attributes:
 .. data:: job_contact 
 
    (Property) Same as attribute :data:`~saga.job.JOB_CONTACT`.
+
+.. autodata:: GENERIC_RESOURCES
+.. data:: generic_resources 
+
+   (Property) Same as attribute :data:`~saga.job.GENERIC_RESOURCES`.
 
 
 .. _job_job:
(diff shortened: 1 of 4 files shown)


andre-merzky:
Dear Rajdeep - thanks for that PR!  That is actually in line with some GPU-related work we are approaching as well.

Still, I am hesitant to merge for two reasons: it changes the API,  and it is slurm specific (or I at least don't see how the GRES format translates to other backends).  Assume that we need similar semantics for, say LoadLeveler, then we would end up adding another attribute (LL-specific then), thus blowing up the API and make it dependent on the chosen backend, but that is precisely what SAGA tries to avoid.

I'll take this up with our group over the next days, and will get back to you with a proposal on how to accommodate the use case w/o breaking the abstraction, if that's ok for you.

Thanks again, Andre.

RajdeepKonwar:
Thank you for your comment Andre! What you are saying makes perfect sense to me.

I wanted to use SAGA for P100 gpu nodes by saying `#SBATCH --gres=gpu:p100:4`. It would be great if you could provide me with some workaround to get this use-case working on the current SAGA release.

Thanks,
Rajdeep

andre-merzky:
Rajdeep, please have a look at the `feature/GPU` branch.  That implements `gres` support for Torque and PBS, by adding a `gres` element to the query part of the `job.service` URL.  I actually think that your approach of adding a job description attribute is much more appropriate, as (a) the parameter is needed on a per-job basis, and (b) it does not make sense to treat the URL query element as a kitchen sink, this is what we have at the moment.

Is that an approach what you could live with for a while, or does that break your abstractions?  Like, that approach is really not useful when submitting many jobs with varying GPU requirements...

What I would *like* to end up with is a job description which specifies something like:

```
jd = saga.job.Description()
jd.cpus = 2
jd.gpus = 1
...
```
which then gets translated into the respective batch directives by the adaptor.  Is that something which would address your use case?  Or do you need to explicitly expose the notion of a node?

Let me try to write up the GPU (and related) requirements for our own use cases over the weekend.  I have been pushing this topic away for too long I think,  so thanks for your prompt to address it :-)

RajdeepKonwar:
Hi Andre, I apologize for the delay in my answer. For now I can work with a local build of my gres implementation. I do like your approach to keep the job descriptions separate for the cpus and gpus. How do you plan to implement the gres option with this approach?

Thanks again for your active response and quick plan of action!

andre-merzky:
yes, proper gres support will  come with the gpu spec.  For now we support gres via URL parameters to the job service URL - ugly, but compliant.  I'll close this PR for now, sorry :/

But please open a ticket if the gres URL support does not work for you!

PR #653, opened 2018-01-20
fix task container get_state reval

andre-merzky:
I am not sure how that went undetected for so long, but here we are...


andre-merzky:
thanks :-)

PR #654, opened 2018-01-29
accept SID settings from upper layers (RP)

andre-merzky:
Matteo - I forgot to assign this one to you, too -- its the counterpart to radical-cybertools/radical.pilot/pull/1539 on the SAGA layer, which now accepts the uid to be specified on session creation, and only throws the dices for a new ID if none is given.

This needs to be merged before radical-cybertools/radical.pilot/pull/1539 - but in its self its backward compatible (the uid parameter is optional).  It is in sync with all other Session constructors in the RCT stack.

### DIFF
--- a/src/saga/base.py
+++ b/src/saga/base.py
@@ -28,13 +28,16 @@ class SimpleBase (object) :
     #
     @rus.takes   ('SimpleBase')
     @rus.returns (rus.nothing)
-    def __init__  (self) :
+    def __init__  (self, uid=None):
 
         if  not hasattr (self, '_apitype') :
             self._apitype = self._get_apitype ()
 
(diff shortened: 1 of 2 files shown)


andre-merzky:
Stack consistency would actually mean `._uid` instead of `._id`, and is something which I would like to see consistent, too.  Alas, that touches too many places in RS, so that would mean a separate branch and PR.

Thanks!

mturilli:
Maybe worth of a ticket?

andre-merzky:
We actually have a catch-all (and thus useless) ticket for consistency (radical-cybertools/radical.utils/issues/8) - I added a note there.  At some point this ticket needs to be split into doable actionables...

PR #658, opened 2018-02-13
Feature/cheyenne

vivek-bala:


src/saga/adaptors/pbspro/pbsprojob.py
```diff
@@ -826,7 +831,7 @@ def _job_get_info(self, job_id, reconnect):
 
         # run the PBS 'qstat' command to get some infos about our job
         # TODO: create a PBSPRO/TORQUE flag once
-        if 'PBSPro_1' in self._commands['qstat']['version']:
+        if 'PBSPro_1' in self._commands['qstat']['version'] or '14.2.4' in self._commands['qstat']['version']:
```
andre-merzky: Would it make sense to check for `14.2` so that it does not break on the next subminor update?

vivek-bala: Yea, I agree. Making the change and sending PR to master.

vivek-bala:
Hey Andre, I tested the changes and changed the base to master. Please review and push a hotfix release.

src/saga/adaptors/pbspro/pbsprojob.py
```diff
@@ -312,6 +312,12 @@ def _pbscript_generator(url, logger, jd, ppn, gres, pbs_version, is_cray=False,
     elif 'PBSPro_13' in pbs_version:
         logger.info("Using PBSPro 13 notation '#PBS -l select=XX' ")
         pbs_params += "#PBS -l select=%d\n" % (nnodes)
+    elif '14.2' in pbs_version:
```
andre-merzky: Is cheyenne using 14.2?  If so, you probably want to check for the hostname first, and then for the pbs version?

vivek-bala: Why 'and' specifically? What if there is another machine that uses 14.2, the condition will fail.

andre-merzky: If cheyenne is using 14.2, then the cheyenne elif branch in this if-tree will never be reached, because this one applies first.

vivek-bala: Ah you're right. Fixed now. 

Question: Do you know why SAGA_PPN gets the value of 72 in the case of cheyenne? ``ppn`` is actually 36 on Cheyenne. I am asking since (1) that is the only difference between the two elif conditions and (2) I am not yet sure if that will break anything in RP.

vivek-bala:
I think this question might have got hidden in last commit:

Question: Do you know why SAGA_PPN gets the value of 72 in the case of cheyenne? ppn is actually 36 on Cheyenne. I am asking since (1) that is the only difference between the two elif conditions and (2) I am not yet sure if that will break anything in RP.

mturilli:
One thing that comes to mind to check is whether that could be the value of logical cores as opposed to physical ones. The fact that you get double the amount of expected cores makes me think about hyperthreading. This may confirm this intuition: https://www2.cisl.ucar.edu/resources/computational-systems/cheyenne/running-jobs/hyper-threading-cheyenne

andre-merzky:
Vivek, can you please reconfirm that this is in a workable state for Cheyennne?  I'll push out a release if that's the case.  Thanks!

vivek-bala:
Testing the Cheyenne stack. Will keep you posted

src/saga/adaptors/pbspro/pbsprojob.py
```diff
@@ -851,11 +781,12 @@ def _job_get_info(self, job_id, reconnect):
 
         # run the PBS 'qstat' command to get some infos about our job
         # TODO: create a PBSPRO/TORQUE flag once
-        if 'PBSPro_1' in self._commands['qstat']['version']:
-            qstat_flag = '-fx'
-        else:
-            qstat_flag ='-f1'
-            
+
+        pbs_version = self._commands['qstat']['version'],
```
vivek-bala: stray comma

src/saga/adaptors/pbspro/pbsprojob.py
```diff
@@ -851,11 +781,12 @@ def _job_get_info(self, job_id, reconnect):
...
+        if   '14.2'     in pbs_version: qstat_flag = '-fx'  # Cheyenne
```
vivek-bala: pbs version is 18.2.<some sub version number> currently. We need to check for '18.2' instead of '14.2'

PR #681, opened 2018-06-08
Fix #680 - Corrected Error on LSF: module object has no attribute Event.

### LINKED ISSUE #680: SAGA LSF might used deprecated code
In lsfjob adaptor, a line reference wrong module attributes and generate this error:
module object has no attribute Event.

marcoooo:


andre-merzky:
Thanks  for catching that! :-)

PR #685, opened 2018-08-30
Feature/summit support

iparask:
This Pull Request is to make sure that the changes in the LSF adaptor are per coding styles and semantics


src/saga/adaptors/lsf/lsfjob.py
```diff
@@ -187,8 +187,13 @@ def _lsfcript_generator(url, logger, jd, ppn, lsf_version, queue, span):
...
         lsf_params += "#BSUB -q %s \n" % queue
 
-    if jd.project is not None:
+    if jd.project is not None and ':' in jd.project:
+        account, reservation = jd.project.split(':',1)
+        lsf_params += "#BSUB -P %s \n" % str(account)
+        lsf_params += "#BSUB -U %s \n" % str(reservation)
+    elif jd.project is not None:
         lsf_params += "#BSUB -P %s \n" % str(jd.project)
+
```
andre-merzky: This is nicely in sync with the slurm impl...

src/saga/adaptors/lsf/lsfjob.py
```diff
@@ -198,6 +203,14 @@ def _lsfcript_generator(url, logger, jd, ppn, lsf_version, queue, span):
...
     lsf_params += "#BSUB -n %s \n" % str(jd.total_cpu_count)
 
+    if jd.total_node_count is None:
+        jd.total_node_count = 1
+
+    lsf_params += "#BSUB -nnodes %s \n" %str(jd.total_node_count)
+
+    if jd.alloc_flags is not None:
+        lsf_params += "#BSUB -alloc_flags %s \n" % ', '.join(jd.alloc_flags)
+
```
andre-merzky: Can you please explain what the semantics of this alloc flag is on summit?

src/saga/adaptors/lsf/lsfjob.py
```diff
@@ -243,7 +256,9 @@ def _lsfcript_generator(url, logger, jd, ppn, lsf_version, queue, span):
                           saga.job.WORKING_DIRECTORY,
                           saga.job.SPMD_VARIATION, # TODO: 'hot'-fix for BigJob
                           saga.job.PROCESSES_PER_HOST,
-                          saga.job.TOTAL_CPU_COUNT],
+                          saga.job.TOTAL_CPU_COUNT,
+                          saga.job.TOTAL_NODE_COUNT,
```
andre-merzky: For other adaptors we use the cores per node count to calculate the number of nodes needed (cobalt, torque).  This means we have to configure or hardcode that value - but I'd rater see that in sync with the other adaptors than changing the API.  For the SAGA layer, API uniformity over different backends is one of the defining features.  Would that be possible to approach in the same way as in those adaptors?

tests/adaptors/lsf/test_lsf_adaptor.py
```diff
@@ -0,0 +1,49 @@
...
+        self._jd.output = 'output.log'
+        self._jd.error = 'error.log'
+        self._jd.queue = 'normal-queue'
+        self._jd.project = 'TestProject'
+        self._jd.wall_time_limit = 70
+        self._jd.total_cpu_count = 24
+        self._jd.total_node_count = 2
+        self._jd.alloc_flags = ['gpumps', 'smt4']
+        
+        self._script = "\n#!/bin/bash \n#BSUB -J Test \n#BSUB -o output.log \n#BSUB -e error.log \n#BSUB -W 1:10 \n#BSUB -q normal-queue \n#BSUB -P TestProject \n#BSUB -n 24 \n#BSUB -nnodes 2 \n#BSUB -alloc_flags gpumps, smt4 \n\nexport  test_env=15 \n/bin/sleep 60 "
```
andre-merzky: Uh, this is a veeeery long line... ;-)

src/saga/adaptors/lsf/lsfjob.py
iparask: https://www.olcf.ornl.gov/for-users/system-user-guides/summit/running-jobs/#common-bsub-options

src/saga/adaptors/lsf/lsfjob.py
iparask: So `processes_per_host` is the job description entry that allows what you are referring to?

Is this because some resources require to tell them how many MPI processes are going to be submitted per node, regardless if they are going to launch any number of threads?

andre-merzky: Well, ideally we would get the number of cores/node from runtime inspection, like [here](https://github.com/radical-cybertools/saga-python/blob/devel/src/saga/adaptors/slurm/slurm_job.py#L379).  Alternatively we can hardcode that value in the adaptor, similar to [here](https://github.com/radical-cybertools/saga-python/blob/devel/src/saga/adaptors/torque/torquejob.py#L312) - we hardcode many things already, and should eventually switch to resource configs on the SAGA layer - but that's a different topic altogether.   I would  like to avoid a situation where the users need to know the core count in order to be able to request a number of cores, so a fallback value for when PPN is not set would really be useful.

src/saga/adaptors/lsf/lsfjob.py
vivek-bala: Should this be:
```
lsf_params += '#BSUB -alloc_flags "%s" \n' % ' '.join(jd.alloc_flags)
```

Note the quotation and separation by space above. The example in the referenced url specifies the flag as:
```
#BSUB -alloc_flags "gpumps smt1"
```

iparask: You are right. For some reason I saw a comma

src/saga/adaptors/lsf/lsfjob.py
iparask: Okay, I see your point. I will add a Summit specific hard coded value for now and mark it with a `TODO`.

src/saga/adaptors/lsf/lsfjob.py
```diff
@@ -187,22 +187,40 @@ def _lsfcript_generator(url, logger, jd, ppn, lsf_version, queue, span):
...
         lsf_params += "#BSUB -u %s \n" % str(jd.job_contact)
 
     # if total_cpu_count is not defined, we assume 1
     if jd.total_cpu_count is None:
-        jd.total_cpu_count = 1
+        total_cpu_count = 1
+    else:
+        total_cpu_count = jd.total_cpu_count
 
     lsf_params += "#BSUB -n %s \n" % str(jd.total_cpu_count)
```
vivek-bala: This causes an error on Summit since '-n' on Summit "Submits a parallel job and specifies the number of tasks in
the job." (source: 'man BSUB').

iparask:
@andre-merzky, I am dismissing the last review, since there are a lot of changes. Please review again and we can see what to do from there.

src/saga/adaptors/lsf_summit/lsfjob_summit.py
```diff
@@ -0,0 +1,1068 @@
...
+        total_cpu_count = jd.total_cpu_count
+
+    ppn = 42
+    number_of_nodes = int(total_cpu_count / ppn)
+    if total_cpu_count % ppn > 0:
+        number_of_nodes += 1
+    lsf_params += "#BSUB -nnodes %s \n" %str(number_of_nodes)
+
+    lsf_params += "#BSUB -alloc_flags 'gpumps smt4' \n"
+
```
andre-merzky: The 5 lines above are the only change really compared to the original adaptor, right?  If that's the case, we should `ifdef` that part and merge back to the original adaptor as soon as it works...

src/saga/adaptors/lsf_summit/lsfjob_summit.py
```diff
@@ -0,0 +1,1068 @@
...
+# --------------------------------------------------------------------
+# some private defs
+#
+_PTY_TIMEOUT = 2.0
+
+# --------------------------------------------------------------------
+# the adaptor name
+#
+_ADAPTOR_NAME          = "saga.adaptor.lsfjob_summit"
+_ADAPTOR_SCHEMAS       = ["lsf", "lsf+ssh", "lsf+gsissh"]
```
andre-merzky: This will conflict with the other LSF adaptor - adaptor selection is based on URLs.

src/saga/adaptors/lsf_summit/lsfjob_summit.py
iparask: As soon as @vivek-bala says it submits jobs correctly, I will put it in the original with a `if summit` statement.

src/saga/adaptors/lsf_summit/lsfjob_summit.py
iparask: Yes, you are right.

iparask:
It passed my test. The test is a unit test not an integration test. I think we need an integration test to address your last comment.

iparask:
I think it is ready. I executed correctly on SummitDev. Summit is still down I think. Nonetheless, I am not sure at this point, how to conduct more tests. Please review so that we can finalize how this adaptor looks like.

tests/adaptors/lsf_summit/test_lsf_adaptor.py
```diff
@@ -0,0 +1,60 @@
...
+        self._jd.executable = '/bin/sleep'
+        self._jd.arguments = 60
+        self._jd.environment = {'test_env': 15}
+        self._jd.output = 'output.log'
+        self._jd.error = 'error.log'
+        self._jd.queue = 'normal-queue'
+        self._jd.project = 'TestProject'
+        self._jd.wall_time_limit = 70
+        self._jd.total_cpu_count = 65
+        self._jd.alloc_flags = ['gpumps', 'smt4']
```
andre-merzky: This is not supported anymore, right?

andre-merzky:
Closed in favor of #700

PR #687, opened 2018-09-24
Fix/issue 686

iparask:
I would like to note that SLURM version on Stampede2 is `18.08.0` and also it return the following as `ppn_vals`:
```
2018-09-24 15:32:36,175: radical.saga.cpi    : MainProcess                     : MainThread     : INFO    :  === ppn_vals: ['272', '96']
```
As a result the check has changed from equal to 1 to larger of equal to 1


andre-merzky:
If stampede2 reports 2 different PPN values, then how do we know which is the 'right' one?  Can we select for a specific node type via some `qsub` flags?

iparask:
There are the `KNL` and `SKX` nodes. The `KNL` nodes have 272 cores, while the `SKX` have 96 logical cores. You can find more [here](https://portal.tacc.utexas.edu/user-guides/stampede2#knl-compute-nodes)

I am open to suggestions.

PR #689, opened 2018-11-12
Pushing version check

### LINKED ISSUE #688: SAGA does not create `-N` flag on SLURM
The `-N` flag is not set on the SLURM adaptor all the time. It is only setup for version `17.11.5`. I propose to setup it up for any version that is larger than that.

This is triggered by [radical.pilot issue 1754](https://github.com/radical-cybertools/radical.pilot/issues/1754).

@andre-merzky, do you agree?

iparask:
This fixes #688

### DIFF
--- a/src/saga/adaptors/slurm/slurm_job.py
+++ b/src/saga/adaptors/slurm/slurm_job.py
@@ -20,6 +20,8 @@ import math
 import time
 import tempfile
 
+from distutils.version import StrictVersion
+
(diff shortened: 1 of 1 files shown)


andre-merzky:
Thanks a lot, Ioannis!  We probably should switch ti StrictVersion checks in several places.  Well, I am sure future tickets will provide that opportunity ;-)

PR #690, opened 2018-11-13
Adding git error in setup.py

iparask:
This change  is due to an error we got from Conda Forge Circle CI.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -79,6 +79,7 @@ def get_version (mod_root):
 
         if  p.returncode   !=  0  or \
             version_detail == '@' or \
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -79,6 +79,7 @@ def get_version (mod_root):
 
         if  p.returncode   !=  0  or \
             version_detail == '@' or \
+            'git-error-while-loading' in version_detail or \
```
andre-merzky: maybe make that a `git-error` check, so that other git errors are also captured?

iparask: Done

PR #691, opened 2018-11-16
Correcting type that was missed

iparask:
We missed this Andre. If this is released let's move forward and do a hotfix. Do you agree?

### DIFF
--- a/src/saga/adaptors/slurm/slurm_job.py
+++ b/src/saga/adaptors/slurm/slurm_job.py
@@ -374,7 +374,7 @@ class SLURMJobService (saga.adaptors.cpi.job.Service) :
                                self.rm.detected_username)
 
         _, out, _ = self.shell.run_sync('scontrol --version')
-        self._version = StrictVesrion(out.split()[1].strip())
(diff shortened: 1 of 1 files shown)


src/saga/adaptors/slurm/slurm_job.py
```diff
@@ -387,7 +387,7 @@ def _open(self):
                     '| xargs echo'
         _, out, _ = self.shell.run_sync(ppn_cmd)
         ppn_vals  = [o.strip() for o in out.split() if o.strip()]
-        if len(ppn_vals) == 1: self._ppn = int(ppn_vals[0])
+        if len(ppn_vals) >= 1: self._ppn = int(ppn_vals[0])
```
iparask: Many systems that have SLURM versions greater that 17.11.5 return multiple values as `ppn`. This is not the best fix, but it selects the first of the returned values.

The issue here is that these systems provide different number of cores per node and setting `_ppn` to the first value will definitely cause, like asking for more nodes as the required

andre-merzky:
Looks good to me - thanks.

PR #695, opened 2018-12-05
fixing SBATCH GPU requirement for Bridges.

iparask:
I do not know how it disappeared.


iparask:
Apparently, it was never merged to devel or master. There is another branch for it called `fix/issue_679`. This PR does the same change.

andre-merzky:
Thanks!

PR #696, opened 2018-12-06
Fix/issue 679

andre-merzky:
This should have been merged already... :/  It add GPU support for Cori, and supports ngpus!=2 on bridges.


iparask:
It is fixed by the other PR. I do not think it is needed anymore

andre-merzky:
Can you point me to that other PR?  Thanks!

Issue #692, opened 2018-11-16
Auto ppn set up might get the wrong value in SLURM adaptor for versions later than 17.11.5


iparask:
On Bridges for example we have SLURM 17.11.7 and it offers at least 2 different core per node counts. 
```
[paraskev@br006 ~]$ scontrol --version
slurm 17.11.7
```

That is 28 for the RM queue and 32 for the GPU queue when the nodes with the P100 GPUs are used.

[Lines 389-391](https://github.com/radical-cybertools/saga-python/blob/devel/src/saga/adaptors/slurm/slurm_job.py#L389-L391) will select 28 cores per node because that is the output of [line 388](https://github.com/radical-cybertools/saga-python/blob/devel/src/saga/adaptors/slurm/slurm_job.py#L388)

```
[paraskev@br006 ~]$ scontrol show nodes | grep CPUTot| sed -e 's/.*\(CPUTot=[0-9]*\).*/\1/g'| sort | uniq -c | cut -f 2 -d = | xargs echo
28 288 32 352 40 64 80 96
```

We need to come up with a way to select the correct value. In addition, Bridges does not require the `-N` flag in the SLURM script, while Stampede2 with SLURM 18.08.3 does.

iparask:
Hey @andre-merzky , do you remember how we said we are going to tackle this? I remember I would pick it up, but I missed it in my todo list

iparask:
It took me a second, but I remember it! 

The solution was to execute `scontrol show partitions | grep -E 'PartitionName|TotalCPUs|TotalNodes'` instead of the command executed now. 

On Bridges this returns:
```
PartitionName=RM
   State=UP TotalCPUs=20160 TotalNodes=720 SelectTypeParameters=NONE
PartitionName=RM-shared
   State=UP TotalCPUs=1932 TotalNodes=69 SelectTypeParameters=NONE
PartitionName=RM-small
   State=UP TotalCPUs=140 TotalNodes=5 SelectTypeParameters=NONE
PartitionName=GPU
   State=UP TotalCPUs=1344 TotalNodes=44 SelectTypeParameters=NONE
PartitionName=GPU-shared
   State=UP TotalCPUs=700 TotalNodes=23 SelectTypeParameters=NONE
PartitionName=GPU-small
   State=UP TotalCPUs=128 TotalNodes=4 SelectTypeParameters=NONE
PartitionName=GPU-AI
[...]

PR #698, opened 2018-12-13
Adding Stampede specific version check only.

iparask:
All other SLURM versions will be treated as before. We need to find a clear way to push cores per node here.

I missed that one. It will fail if the queue the user submits has different cores per node, from the first one. Already stumbled upon it.

This change makes it specific to Stampede2 version, which has the correct number of cores per node in the first value.

### DIFF
--- a/src/saga/adaptors/slurm/slurm_job.py
+++ b/src/saga/adaptors/slurm/slurm_job.py
@@ -20,8 +20,6 @@ import math
 import time
 import tempfile
 
-from distutils.version import StrictVersion
-
 SYNC_CALL  = saga.adaptors.cpi.decorators.SYNC_CALL
 ASYNC_CALL = saga.adaptors.cpi.decorators.ASYNC_CALL
 
@@ -374,7 +372,7 @@ class SLURMJobService (saga.adaptors.cpi.job.Service) :
                                self.rm.detected_username)
 
         _, out, _ = self.shell.run_sync('scontrol --version')
-        self._version = StrictVersion(out.split()[1].strip())
+        self._version = out.split()[1].strip()
         self._logger.info('slurm version: %s' % self._version)
 
         ppn_pat   = '\'s/.*\\(CPUTot=[0-9]*\\).*/\\1/g\'' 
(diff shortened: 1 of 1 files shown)


src/saga/adaptors/slurm/slurm_job.py
```diff
@@ -479,7 +479,7 @@ def _job_run (self, jd) :
             # we start N independent processes
             mpi_cmd = ''
 
-            if self._version >= StrictVersion('17.11.5'):
+            if self._version in [StrictVersion('17.11.5'),StrictVersion('18.08.3')]:
```
andre-merzky: Why did the `>=` did not accomplish the same?  What happens if Stampede is moving to `18.08.4`?  If `StrictVersion` is not working as intended,  we should not use it at all -- it will then only obfuscate the code...

iparask: It did not because we haven't found a way to recognize the correct number of cores per node. 

Specifically, on [slurm_job.py#L388](https://github.com/radical-cybertools/saga-python/blob/fix/slurm_versioning/src/saga/adaptors/slurm/slurm_job.py#L388) we discover the number of cores per node and grab the first number always. Although that is not the case of all queues as #692 discusses. That is not the case for the use cases we support on Stampede2. The first value is the correct one.

I agree that `StrictVersion` is not needed right now, but I think it will be useful when we are able to get the correct information for `cores_per_node`

iparask: I'll remove it. I am not sure that going with `hostnames` will solve our problem. I remember at some point we said that we may want this kind of input to be configurable. It may make more sense to have these values in a json file like:
```json
{"resourse": {"queue1" : cores,
              "queue2" : cores}
....
```

This might make our life easier, especially with queue were the nodes can be different, such the GPU queues of Bridges, where depending on the GPU type we get a different number of cores per node.

andre-merzky: Yes, that will be the right thing to do.  I am working on an overhaul of the SAGA config management (in `feature/json_configs`) - but until then this is what we have *sigh*.  Thanks!

PR #699, opened 2019-02-03
Feature/issue grlsd 95

andre-merzky:
This is confirmed working as per https://github.com/radical-collaboration/extasy-grlsd/issues/95.  But this needs further iteration to be made optional due to:

```sh
$ cat test.pbs
#!/bin/bash

#PBS -q debug
#PBS -A PRAC_BAMM
#PBS -l nodes=2:ppn=32
#PBS -l walltime=0:15:00
#PBS -l flags=commtransparent

aprun -n 1 hostname


$ qsub test.pbs
sh: line 1: 30809 Aborted                 /sw/user/scripts/vfyproj merzky > /tmp/.alloc.merzky.projects.tmp 2> /dev/null

ERROR: Job is not allowed to use 'commtransparent' flag.
Please contact Blue Waters support at help+bw@ncsa.illinois.edu for assistance.
qsub: Your job has been administratively rejected by the queueing system.
qsub: There may be a more detailed explanation prior to this notice.
```


andre-merzky:
The flag is now optional, this should be mergeable

PR #701, opened 2019-02-04
Feature/timing

andre-merzky:
support timing of job submissions


andre-merzky:
Nah, we should not do this, but instead add profile probes...

PR #705, opened 2019-02-27
Change SMT from 1 to 4

iparask:
What is says in the title :)


mturilli:
Uff, that was long and difficult to review. We should really reduce the size of our PRs!

andre-merzky:
> Uff, that was long and difficult to review. We should really reduce the size of our PRs!

I  don't like the PR: no documentation, no design document, no unit tests.  Pffft...

PR #708, opened 2019-03-01
Fix/issue 707

### LINKED ISSUE #707: SAGA to expose environment variable about SMT setting for Summit
Just follow up ticket from an email thread with ORNL. There are no explicit env variables that expose that, we will have to look at the cpuset (at /sys/fs/cgroup/cpuset/cpuset.cpus) and parse it. Easier solution would be to have SAGA expose an env variable.

andre-merzky:
This fixes #707


mturilli:
@iparask To add unit test and review

andre-merzky:
unit test has been fixed.  I made a stupid mistake and merged the SMT semantics into `feature/summit` when cleaning up there, and only realized after several changes - apologies.  :-(

andre-merzky:
The dir rename is done.

andre-merzky:
For the reasoning behind the adaptor rename see discussion in #715

andre-merzky:
This is now implemented in the feature/summit branch.

PR #711, opened 2019-03-22
Document missing path normalization

mturilli:


mturilli:
@andre-merzky can we fix the tests so that they pass in Travis?

andre-merzky:
Thanks for the PR, Matteo!  Alas, it will not really resolve the issue, since the double slashes are often ingested from other pieces of code, and that API path (where the URL elements are set individually) does then not apply.

I tried to provide a code fix in radical-cybertools/radical.utils/pull/159 - can you have a look if that looks ok for you?

mturilli:
Thanks Andre. I left a couple of comments on the PR 159

mturilli:
Fixed in utils instead of documenting the error

PR #715, opened 2019-03-27
Feature/summit

andre-merzky:
The lsfsummit adaptor is now merged into the lsf adaptor, as it is (a) functional on summit and summitdev, and (b) has minimal changes to the lsf adaptor - and those changes are required there anyway.  The only summit specificity is the core count and the default SMT setting - and both will  move into configurations files after 0.60.

This is thus ready for merge.  For the time being, do *not* delete the `feature/summit` branch, until the full RCT stack support for summit is released..


andre-merzky:
The conflicts are resolved - @iparask , looks ok to you?

andre-merzky:
thanks :-)  Will have to wait for summit_am though...

PR #716, opened 2019-04-13
Version/2

andre-merzky:
I would like to sync current v2 with devel.  By now this PR mostly contains PEP-8 fixes and the like.


src/radical/saga/adaptors/irods/irods_replica.py
```diff
@@ -331,41 +340,43 @@ def irods_get_resource_listing(self):
...
                     # ----- is not printed if there are no further entries
                     # so we have to make sure to check we don't pop off
                     # an empty stack too
                     #
-                    # TODO: ACTUALLY SAVE THE LIST OF RESOURCES IN A RESOURCE GROUP
-                    while len(cw_result_list)>0 and (not line.startswith("-----")):
-                        line=cw_result_list.pop(0)
-    
+                    # TODO: SAVE THE LIST OF RESOURCES IN A RESOURCE GROUP
+                    while len(cw_result_list) > 0 and \
```
iparask:
```suggestion
                    while cw_result_list and \
```
I think pep8 does not like checking if lists have elements by a len check.

andre-merzky:
There are many pep-8-ties left, agree - but you are right, I should clean those for the lines touched at least...

andre-merzky:
done

PR #718, opened 2019-05-23
ensure cray detection

andre-merzky:
fix cray detection (failed on titan)


andre-merzky:
This basically ensures that the `aprun` check is done on empty and `False` values for `self.is_cray`.

PR #719, opened 2019-05-28
Hotfix/delay expand

andre-merzky:
This PR accompanies the radical-cybertools/radical.utils/pull/165.  It ensures that environment expansion on config files is done with the environment of the *target* host when connecting an ssh/gsissh job service.

### DIFF
--- a/src/radical/saga/adaptors/base.py
+++ b/src/radical/saga/adaptors/base.py
@@ -28,7 +28,7 @@ class Base(object):
     #
     # FIXME: phase out adaptor_options
     #
-    def __init__ (self, adaptor_info, adaptor_options=None):
+    def __init__ (self, adaptor_info, adaptor_options=None, expand_env=True):
 
         # FIXME: engine is loading cfg already, here we load again...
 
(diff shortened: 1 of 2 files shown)


src/radical/saga/adaptors/base.py
```diff
@@ -40,7 +40,9 @@ def __init__ (self, adaptor_info, adaptor_options=None) :
         self._logger  = ru.Logger('radical.saga.api')
 
 
-        self._cfg     = ru.Config(module='radical.saga', name=self._name)
+        # we need to expand later once we got env from the remote resource
+        self._cfg     = ru.Config(module='radical.saga', name=self._name,
+                                  expand=False)
```
mturilli: As we are changing the base class, are we changing the default expand behavior for all the adaptors?

andre-merzky: Yes, indeed, good point.  The `shell_job` adaptor is the only active adaptor in use which uses that env expansion on the target host.  But let me clean this up for the adaptors which use local expansion.

Thanks.

PR #724, opened 2019-06-24
Hotfix/lsf

### LINKED ISSUE #723: Include TotalGPUCount in lsfjob adaptor
Summit runs fail on Summit with:
```
2019-06-24 10:50:31,384: radical.saga        : pmgr.0000.launching.0           : MainThread     : ERROR   : BadParameter: 'JobDesc
ription.TotalGPUCount' (6) not supported by radical.saga.adaptors.lsfjob
```

andre-merzky:
This fixes #723

### DIFF
--- a/examples/jobs/lsfjob.py
+++ b/examples/jobs/lsfjob.py
@@ -54,6 +54,7 @@ def main():
         jd.arguments         = ['$FILENAME']
 
         jd.total_cpu_count   = 42
+        jd.total_cpu_count   = 4
 
         jd.queue             = "batch"
         jd.project           = "GEO111" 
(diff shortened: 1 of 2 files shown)


examples/jobs/lsfjob.py
```diff
@@ -53,10 +53,11 @@ def main():
...
         jd.arguments         = ['$FILENAME']
 
-        jd.total_cpu_count   = 16
+        jd.total_cpu_count   = 42
+        jd.total_cpu_count   = 4
 
-        jd.queue             = "regular"
-        jd.project           = "URTG0003" 
+        jd.queue             = "batch"
+        jd.project           = "GEO111"
```
mturilli: For the future, we should probably use an env variable to store this?

andre-merzky: We should move to config files.  SAGA supports them now, but they are not yet consistently used in the adaptors...

PR #726, opened 2019-07-03
frontera, here we come!

andre-merzky:


andre-merzky:
@iparask @mturilli  : please review - thanks

src/radical/saga/utils/misc.py
```diff
@@ -43,12 +43,20 @@ def host_is_local (host) :
...
+
+    elif host == 'localhost':
+        return True
+
+    else:
+        sockhost = socket.gethostname()
+        while sockhost:
+            if host == sockhost:
+                return True
+            sockhost = '.'.join(sockhost.split('.')[1:])
```
iparask: I am not sure I understand this loop. You slowly remove parts of the host name until you find the host?

andre-merzky: Yes, indeed - this helps for hostnames like `login1.summit.ornl.gov` or `batch1.summit.ornl.gov`, where we commonly use `summit.ornl.gov` as a hostname which is dynamically proxied to one of the candidate nodes for load balancing.  That is a common scheme on our machines.  On those machines the canonical hostname would not match, but removing the  leading parts help...

PR #728, opened 2019-07-06
Rc/v0.70.0

andre-merzky:


andre-merzky:
I should add that it also takes a long time on travis.

Issue #735, opened 2019-08-29
Python3 devel error in attributes


iparask:
While running a RADICAL Pilot example I got the following error:
```
  File "/home/iparask/anaconda3/envs/python3.5/lib/python3.5/site-packages/radical/pilot/session.py", line 384, in _load_resource_configs
    rcs = ResourceConfig.from_file(config_file)
  File "/home/iparask/anaconda3/envs/python3.5/lib/python3.5/site-packages/radical/pilot/resource_config.py", line 200, in from_file
    rcfgs[label] = ResourceConfig(label, cfg)
  File "/home/iparask/anaconda3/envs/python3.5/lib/python3.5/site-packages/radical/pilot/resource_config.py", line 218, in __init__
    attributes.Attributes.__init__(self, seeding_dict)
  File "/home/iparask/anaconda3/envs/python3.5/lib/python3.5/site-packages/radical/saga/attributes.py", line 340, in __init__
    us_key = self._attributes_t_underscore(key)
  File "/home/iparask/anaconda3/envs/python3.5/lib/python3.5/site-packages/radical/saga/attributes.py", line 744, in _attributes_t_underscore
    temp = Attributes._camel_case_regex_1.sub(r'\1_\2', key)
TypeError: cannot use a string pattern on a bytes-like object
```

andre-merzky:
Ah, it seems I am reading things out of order - apologies. Either way, shouldn't we rather make sure that  what we pass to the attribute layer is not bytes but strings?  The attribute API requires to pass strings, really.  I think the underlying problem is that `ResourceConfig.from_file()` gets data passed from `ru,read_json()` as byte strings.  I think the best approach for our stack would be to make sure that all data returned from `read_json()` are ASCII encoded, which saves us from doing conversions all over the place whenever we need to use some piece of data.  What do you think?

mturilli:
We discussed the general version of this issue at our last developer meeting and we all agree we should have a 'principled' approach to managing byte/str type in p3. I agree we should assume that data returned from read_jason() is always ascii encoded and we should have a similar approach for all the other cases we find.

andre-merzky:
This should be fixed now - the config file parser converts all data from bytes into their respective data types.

iparask:
I think so as well. I'm closing it

PR #736, opened 2019-08-29
fixing attributes bytes to str error

iparask:
This PR fixes issue #735

### DIFF
--- a/src/radical/saga/attributes.py
+++ b/src/radical/saga/attributes.py
@@ -741,7 +741,7 @@ class Attributes (_AttributesBase, ru.DictMixin) :
 
 
         if  force or d['camelcasing'] :
-            temp = Attributes._camel_case_regex_1.sub(r'\1_\2', key)
+            temp = Attributes._camel_case_regex_1.sub(r'\1_\2', str(key))
             return Attributes._camel_case_regex_2.sub(r'\1_\2', temp).lower()
         else :
             return key


src/radical/saga/attributes.py
```diff
@@ -741,7 +741,7 @@ def _attributes_t_underscore (self, key, force=False) :
 
 
         if  force or d['camelcasing'] :
-            temp = Attributes._camel_case_regex_1.sub(r'\1_\2', key)
+            temp = Attributes._camel_case_regex_1.sub(r'\1_\2', str(key))
             return Attributes._camel_case_regex_2.sub(r'\1_\2', temp).lower()
```
andre-merzky: Where was this an error?  And why only on this call? Should we re-enable the type checks (see line 726)?  But they are costly...

iparask: Are type casting costly in Python 3? If not then we can keep it to be sure that it is a string.

andre-merzky: It's not about cost - but if we do it this way, we will be chasing after this problem endlessly.  Now you provided a fix for this call - but the next user might use the dict key in another call.  Or in another class.  We will have to sprinkle `str()` all over the place.  I mean, we can do that, but then we'd rather do it coordinated like:
```
def some_method(key, size):
   key=str(key)
   size=int(size)
   ...
```
How do you handle cases where the type can be more than one?
```
def some_method(key, size, data=None):
    ...
    if not ininstance(data, type(None), int, basestring):
       data = str(data)
```
Then we can just as well use the type check decorators again.
It is not that your fix won't work - it will.  But it does not remove the cause of the problem, only the effect in a specific case..

andre-merzky:
This is now fixed centrally in the  RU json parser.

PR #740, opened 2019-10-08
fix based on #739

### LINKED ISSUE #739: PTY shell factory compares a str with float
I got this error when I tried to run an RP example:

```
2019-10-08 04:14:27,307: radical.saga.pty    : MainProcess                     : MainThread     : ERROR   : unorderable types: float() > str()
  File "/home/iparask/test_rp/lib/python3.5/site-packages/radical/saga/utils/pty_shell_factory.py", line 292, in _initialize_pty
    if time.time() - time_start > timeout:
TypeError: unorderable types: float() > str()
```

iparask:
Fixes #739

### DIFF
--- a/src/radical/saga/utils/pty_shell_factory.py
+++ b/src/radical/saga/utils/pty_shell_factory.py
@@ -231,7 +231,7 @@ class PTYShellFactory (object, metaclass=ru.Singleton) :
             prompt     = info['prompt']
             logger     = info['logger']
             latency    = info['latency']
-            timeout    = info['ssh_timeout']
+            timeout    = float(info['ssh_timeout'])
 
             pty_shell.latency = latency
 


src/radical/saga/utils/pty_shell_factory.py
```diff
@@ -231,7 +231,7 @@ def _initialize_pty (self, pty_shell, info, posix=None) :
             prompt     = info['prompt']
             logger     = info['logger']
             latency    = info['latency']
-            timeout    = info['ssh_timeout']
+            timeout    = float(info['ssh_timeout'])
```
andre-merzky: we then should also cast `latency`?  Any idea why that suddenly is a string?  Otherwise this looks good.

PR #751, opened 2019-12-02
Feature/proc mgmt

andre-merzky:
I forgot to open this PR also :-/  It contains some collateral changes caused by removal of `ru.Thread`s and `ru.Process`es, and by the changes in the config management.  Plus some (partial) code cleanup in the touched files.


andre-merzky:
Tests are passing locally now - but pep8 a nd pylint will take a while :-(

src/radical/saga/adaptors/base.py
```diff
@@ -51,43 +53,45 @@ def __init__ (self, adaptor_info, adaptor_options=None, expand_env=True):
...
-                         % self._name)
+    def sanity_check(self):
+        '''
+        This method can be overloaded by adaptors to check runtime conditions on
+        adaptor load time.  The adaptor should raise an exception if it will not
+        be able to function properly in the given environment, e.g. due to
+        missing dependencies etc.
+        '''
+
+        pass
```
iparask: The doc string is not in sync with the code. It says that an exception is raised, but the method just passes.

src/radical/saga/adaptors/srm/srmfile.py
```diff
@@ -84,7 +84,7 @@ class Adaptor(rsab.Base):
     def __init__(self) :
 
         rsab.Base.__init__(self, _ADAPTOR_INFO, _ADAPTOR_OPTIONS)
-        self.pty_url = self._cfg['pty_url']
+        self.pty_url = self._cfg.get('pty_url', 'fork://localhost/')
```
iparask: Is there a reason you want the pty_url to default to localhost?

src/radical/saga/engine/engine.py
```diff
@@ -251,7 +252,8 @@ def _load_adaptors (self, inject_registry=None):
             adaptor_enabled = False
 
             try :
-                adaptor_config  = ru.Config('radical.saga', name=adaptor_name)
+                adaptor_config  = ru.Config('radical.saga.adaptors',
+                                            name=adaptor_name)
```
iparask: `'radical.saga.adaptors'` is the module right? May I suggest to change it to:
```python
adaptor_config = ru.Config(module='radical.saga.adaptor',
                                                      name=adaptor_name)
```

It creates a redundancy, but I think it makes more consistent with lines 132-134

src/radical/saga/utils/pty_process.py
```diff
@@ -97,16 +98,22 @@ def __init__ (self, command, cfg=None, logger=None) :
...
         self.logger.debug ("PTYProcess init %s" % self)
 
 
+        name = None
+        if isinstance(cfg, str):
+            name = cfg
+            cfg  = None
+
+        self.cfg = ru.Config('radical.saga.session', name=name, cfg=cfg)
+        self.cfg = self.cfg.pty
```
iparask: I see this as a dependency loop. `name` is set as the cfg. The `cfg` is then setup as an `ru.Config`, which the is changed to a specific entry of the cfg dict. Why all these? Would it make more sense to have the name as a constructor parameter, as well as, the config? Name can be `'utils'` and cfg can be `None` by default.

What do you think?

src/radical/saga/utils/pty_shell.py
```diff
@@ -212,11 +212,12 @@ def __init__ (self, url, session=None, logger=None, opts=None, posix=True,
...
-
-        # opts passed on construction overwrite file config
-        if opts:
-            self.cfg = ru.dict_merge(self.cfg, opts, policy='overwrite')
+        name = None
+        if isinstance(cfg, str):
+            name = cfg
+            cfg  = None
+        self.cfg = ru.Config('radical.saga.session', name=name, cfg=cfg)
+        self.cfg = self.cfg.pty
```
iparask: Same as before

src/radical/saga/utils/pty_shell_factory.py
```diff
@@ -143,12 +144,19 @@ class PTYShellFactory (object, metaclass=ru.Singleton) :
...
+        self.rlock      = mt.RLock()
+        self.registry   = dict()
+
+        name = None
+        if isinstance(cfg, str):
+            name = cfg
+            cfg  = None
+
+        self.cfg = ru.Config('radical.saga.session', name=name, cfg=cfg)
+        self.cfg = self.cfg['pty']
```
iparask:
```suggestion
        self.cfg = self.cfg.pty
```

src/radical/saga/utils/pty_shell_factory.py
```diff
@@ -143,12 +144,19 @@ class PTYShellFactory (object, metaclass=ru.Singleton) :
...
-        self.cfg        = ru.Config('radical.saga', 'utils')['pty']
+        self.rlock      = mt.RLock()
+        self.registry   = dict()
+
+        name = None
+        if isinstance(cfg, str):
+            name = cfg
+            cfg  = None
+
+        self.cfg = ru.Config('radical.saga.session', name=name, cfg=cfg)
```
iparask: Same as before

src/radical/saga/adaptors/base.py
andre-merzky: It is in sync ;-)  This is the base class' method - it _can_ be overloaded to raise an exception on check errors.  In fact, it did not adhere to the description before the change...

src/radical/saga/adaptors/srm/srmfile.py
andre-merzky: because that is the documented default value - see line 27.

src/radical/saga/engine/engine.py
andre-merzky: I removed the `module=` part in 132-134, if that's ok for you?  Rationale being that `module` is actually not a good name, it's more a namespace, really, and we may want to change that on RU level (out of scope for this PR though).  Does that work?

src/radical/saga/utils/pty_process.py
andre-merzky: That would make perfect sense, and I pondered doing so, but decided to keep the API stable and overload the `cfg` parameter.  We do the same (for good or for worse) in other places in RP, where a `cfg` can be an explicit cnofig dictionary or `ru.Config` instance, or can be a string which points to a know config.  I agree that we may want to clean this up at some point, but I'm not sure if we should right now?

andre-merzky:
> I realized that in some files the docstring was changed from """ to ''', but not in all files. I mention it because it was a very specific change.

That's part of the cleanup, and I apply it whenever I stumble over it.  But since the `"""` version is not a PEP violation, but only of our own guidelines, I did not yet actively try to do a search and replace.

src/radical/saga/utils/pty_process.py
mturilli: I would like to merge this ASAP as it is now possibly blocking a user (Eugene). Could you open a ticket about the cleanup so that it stays on our radar/roadmap?

src/radical/saga/engine/engine.py
iparask: Yes!

src/radical/saga/utils/pty_process.py
iparask: I agree with @mturilli and open an issue for this.

PR #762, opened 2020-02-03
Verify logger instance before use

### LINKED ISSUE #757: Uninitialised logger causes crash in pbsprojob,py
I'm having a go at using radical-saga, installed via pip. I'm trying to submit jobs to a PBSPro queue. I'm getting the following exception in pbsprojob.py:
```
line 160, in _to_saga_jobstate
    logger.debug('check state: %s', job_state)
AttributeError: 'NoneType' object has no attribute 'debug'
```
In the function declaration I can see the logger is defaulted to None:

`def _to_saga_jobstate(job_state, retcode, logger=None)`

It's being called from _job_get_info, but the logger isn't being passed to the function:

`job_info['state'] = _to_saga_jobstate(job_state, retcode)`

I've tried changing this explicitly pass the logger:

`job_info['state'] = _to_saga_jobstate(job_state, retcode, self.logger)`

However, when I try to later query a job's state via `job_service.get_job`, the logger hasn't been instantiated in the object and another crash happens.
To get round this I've had to change the function call back to what it was, and comment out both the logger calls in `_to_saga_job_state`

andre-merzky:
This fixes #757

### DIFF
--- a/src/radical/saga/adaptors/pbspro/pbsprojob.py
+++ b/src/radical/saga/adaptors/pbspro/pbsprojob.py
@@ -9,7 +9,7 @@ __license__   = "MIT"
 
 
 import re
-import os 
+import os
 import time
 import datetime
 import threading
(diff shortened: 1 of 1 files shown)


andre-merzky:
thanks @AymenFJA !

PR #774, opened 2020-03-22
Hotfix/cobalt

andre-merzky:
revive cobalt support - tested (within bounds)  on Theta


andre-merzky:
@iparask : maybe you have the time to review this one?

src/radical/saga/adaptors/slurm/slurm_job.py
```diff
@@ -404,6 +404,9 @@ def _open(self):
         elif 'frontera' in self.rm.host.lower():
             self._ppn = 56
 
+        elif 'comet' in self.rm.host.lower():
```
iparask, line 407: Why is there a change on slurm? This is not cobalt relevant?

andre-merzky: This is because of an out-of-sync merge, will be resolved in devel.  Thanks for noticing.

andre-merzky:
Thanks @iparask, and fair enough on cobalt access.  I decided to merge this anyway, as we tested during the covid effort, and it's an isolated code change (limited to the adaptor code).

PR #775, opened 2020-03-22
Fix/issue 761

andre-merzky:
simplify exception handling, user exception chaining.


src/radical/saga/adaptors/redis/redis_advert.py
```diff
@@ -232,7 +232,7 @@ def attribute_getter (self, key) :
 
         except Exception as e :
             self._logger.error ("get_key failed: %s" % e)
-            raise e
+            raise
```
iparask, line 235: Wouldn't it make sense to remove the log and keep the exception message in the raise?

src/radical/saga/attributes.py
```diff
@@ -552,9 +552,9 @@ def _attributes_t_call_setter (self, key, val) :
             try :
                 d['attributes'][key]['recursion'] = True
                 key_setter (val)
-            except Exception as e :
+            except:
                 can_ignore -= 1
-                if not can_ignore : raise e
+                if not can_ignore : raise
```
iparask, line 557: Why did you remove the exception message?

src/radical/saga/adaptors/redis/redis_advert.py
andre-merzky: `raise e` does the same as `raise` - a raise w/o explicit exception raised the current exception stack.  We use that syntax elsewhere and it is the recommended syntax (https://www.python.org/dev/peps/pep-3109/), thus the change.

andre-merzky:
> I read the whole PR and I have a couple of questions. I am not sure though how to test these changes.

The changes are really just refactoring, they don't change any semantics - the same exceptions and exception types as before are raised.  I don't think this is testable.


> Also, I realized that the code is not even close to our coding guidelines. May I suggest that in the PRs that are doing code maintenance to report the pylint score before and after? It will really help.

The PR is doing that one refactoring item (clean up exception raising), as we have a ticket for it.  I won't be able to fix linting for all 26 files touched - but I am sure that one specific improvement in many places is better than not doing that improvement in the first place, ain't it? :-)

iparask:
> The PR is doing that one refactoring item (clean up exception raising), as we have a ticket for it. I won't be able to fix linting for all 26 files touched - but I am sure that one specific improvement in many places is better than not doing that improvement in the first place, ain't it? :-)

I agree completely, I'm just suggesting to do locally something like `pylint .....` before and after the changes locally and just include the change in the score. I am not sure if it makes sense, but at least shows the level of improvement. If the change is 0 then it does not really matter. Either way, we can move forward with this PR.

andre-merzky:
```
(ve3)  rivendell  merzky  ~/radical/radical.saga.devel $ git co devel
Switched to branch 'devel'
Your branch is up to date with 'devel'.

(ve3)  rivendell  merzky  ~/radical/radical.saga.devel  [devel] $ pylint src/ 2>&1 | tail -n 2
Your code has been rated at 8.43/10

(ve3)  rivendell  merzky  ~/radical/radical.saga.devel  6  [devel] $ git co fix/issue_761
Switched to branch 'fix/issue_761'
Your branch is up to date with 'origin/fix/issue_761'.

(ve3)  rivendell  merzky  ~/radical/radical.saga.devel  [fix/issue_761] $ pylint src/ 2>&1 | tail -n 2
Your code has been rated at 8.46/10 (previous run: 8.43/10, +0.02)
```
Make of that what you will :-)

PR #788, opened 2020-06-09
replace PIL with Pillow

Abdullah-Ghani:
Use Pillow instead, as PIL is basically dead. Pillow is a maintained fork of PIL.


andre-merzky:
I take your word for it! :-)

andre-merzky:
PS.: please always apply labes to issues and pull requests.  Thanks!

PR #794, opened 2020-07-08
Updated `memory_per_node` parameter

mtitov:


mtitov:
@andre-merzky thank you! the way I was thinking is to keep it where parameter `n_nodes` is defined

andre-merzky:
> @andre-merzky thank you! the way I was thinking is to keep it where parameter `n_nodes` is defined

Yeah, but that parameter is also always defined in the same way ;)

How about:
```
n_nodes = None
if self._ppn:
   n_nodes = ...

if total_memory:
  assert(n_nodes)
  mem_per_node = total_memory / n_nodes
```

Those calculations are, at the end, really independent from the specific target resource, and should also work for resources we don't really check in the code.

Sorry, I should have though about this before, did only so now after you merged, so let's clean this up whenever it becomes a problem.

PR #796, opened 2020-07-09
Update `n_nodes` handling

mtitov:


mtitov:
@andre-merzky shouldn't we let slurm to raise an error if it would cause the issue? if some machine will fall into the option without `n_nodes` (and/or by default it might consider 1 node) and will use that `memory_per_node`/`total_memory` at `--mem` option, thus the restriction will be from our side (kinda ahead)

andre-merzky:
Yeah, I see your point - but (a) errors from Slurm are harder to catch and to report, and (b) the SAGA spec requires to raise an error if a parameter is specified but cannot be interpreted.

mtitov:
thanks @andre-merzky, that makes sense. Updated accordingly

PR #797, opened 2020-07-16
Revert "Fix bridges GPU gres flag"

iparask:
Reverts radical-cybertools/radical.saga#792


iparask:
Sorry @andre-merzky I did not change the target branch to devel from PR 792.

andre-merzky:
Oy, I see... - sorry, I missed that, too.  But I think I can just push a new release.

iparask:
Better not. I would say revert and let me open a PR for devel because it is missing the latest changes. I made a small mess there.

andre-merzky:
*\me tips hat*

PR #798, opened 2020-07-16
Hotfix/bridges gpu

iparask:
This is the correct fix that includes K80 and P100 GPUs.


iparask:
Those are coming from master. I checked after I saw the diff and the changes, apart from GRES, are coming from master. I can recreate the branch from devel and have a smaller diff.

andre-merzky:
Ah, got it - I synced devel now, looks better...

PR #803, opened 2020-08-05
Remove redis-adaptor loading

mtitov:
Should some other adaptors be removed (comment out in config) from the loading as well?

p.s. Redis adaptor is not removed completely, just not loading (if there will be a request for it, then it can be turned on and some it parts should be updated according to the latest version of RU)


mtitov:
@andre-merzky I found this inconsistency (since it was almost at the top, but didn't check everything there, may be it is the only one):
```
import radical.utils.threads as rut
...
rut.Thread.__init__(self, self.work)
```
and `radical.utils.threads` doesn't have `Thread` anymore

PR #805, opened 2020-08-07
"fix" container add

andre-merzky:
At some point, task membership in a list seems to have broken - I have no idea how and why.  Note that this also holds for a plain Python list (so, when disabling the TASK attribute and adding `self.tasks = list()` in `__init__`.  the tasks *do* have different object IDs.  No idea how `in` is implemented (and don't care right now)


mtitov:
for the method `add` ([L346](https://github.com/radical-cybertools/radical.saga/blob/72dcca2dcc36195625be659785a101d96c33635d/src/radical/saga/task.py#L346)) I have the following proposal:
```python
task_in_tasks = next(filter(lambda t: t._id == task._id, self.tasks), None)
if task_in_tasks is None:
    self.tasks.append(task)
```
later there is a similar approach in the method `remove` ([L357](https://github.com/radical-cybertools/radical.saga/blob/72dcca2dcc36195625be659785a101d96c33635d/src/radical/saga/task.py#L357))
```python
task_in_tasks = next(filter(lambda t: t._id == task._id, self.tasks), None)
if task_in_tasks is not None:
    self.tasks.remove(task)  # ! should be `remove` not `delete`
```

BTW later there is a reference to task's id ([L585](https://github.com/radical-cybertools/radical.saga/blob/72dcca2dcc36195625be659785a101d96c33635d/src/radical/saga/task.py#L585)), I think it is a mistake, there should be `t._id` not `t.id`, and the similar approach described above could be applied:
```python
task_in_tasks = next(filter(lambda t: t._id == id, self.tasks), None)
if task_in_tasks is not None:
    return task_in_tasks
```

andre-merzky:
Sure, we can work around this - but I would like to rather understand why the hell this doesn't work anymore.  It for sure *did* work in the past, but I wasn't able to bisect yet (gave up because I was simply too frustrated). 

For the t.id: good catch!  This will only work if the task is a job (`rs.job.Job` inherits from `rs.task.Task`, and it has an `id` attribute, the job id).  I'll add an id to the task (which is useful either way)

mtitov:
we still can keep a note for that, just I don't like idea to add every task, since don't know how duplicates can mess the workflow (+ `remove` method also should be update) p.s. if to add every task, will that `remove` method delete *all* that copies?

src/radical/saga/task.py
```diff
@@ -94,6 +94,11 @@ def __init__ (self, _adaptor, _method_type, _method_context, _ttype) :
         self._attributes_set_getter(c.STATE,     self.get_state)
         self._attributes_set_setter(c.STATE,     self._set_state)
 
+        self._attributes_register  (c.ID,        None,    satt.STRING, satt.SCALAR, satt.READONLY)
```
mtitov, line 97: there is no `ID` in constants, that constant was in `radical.saga.job.constants`

src/radical/saga/task.py
```diff
@@ -121,6 +126,26 @@ def __init__ (self, _adaptor, _method_type, _method_context, _ttype) :
             pass
 
 
+    # --------------------------------------------------------------------------
+    #
+    @rus.takes   ('Task',
+                  rus.optional (rus.one_of (SYNC, ASYNC, TASK)))
```
mtitov, line 132: `SYNC, ASYNC, TASK` -> `c.SYNC, c.ASYNC, c.TASK`

src/radical/saga/task.py
```diff
@@ -121,6 +126,26 @@ def __init__ (self, _adaptor, _method_type, _method_context, _ttype) :
...
+    @rus.returns ((rus.nothing, str, st.Task))
```
mtitov, line 133: this one is tricky, it might be `st.Task` -> `'Task'` (?)

src/radical/saga/task.py
```diff
@@ -121,6 +126,26 @@ def __init__ (self, _adaptor, _method_type, _method_context, _ttype) :
...
+        """
+        get_id()
+
+        Return the task ID.
+        """
+
+        try:
+            id = self._adaptor.get_id (ttype=ttype)
+        except:
+            id = self._uid
```
mtitov, line 144: `id = self._uid` -> `id = self._id`

src/radical/saga/job/job.py
```diff
@@ -144,21 +142,6 @@ def __str__  (self) :
...
-    def get_id   (self, ttype=None) :
-        """
-        get_id()
-
-        Return the job ID.
-        """
-        id = self._adaptor.get_id (ttype=ttype)
-        return id
-
-
```
mtitov, line 161: leftover at L414

mtitov:
May be keep ID at Job level? (it is not that needed at Task level)

andre-merzky:
I did not have to say this for a long time (like, 2 weeks), but: fuck Python.  Exhibit 1234:

compare
```py
#!/usr/bin/env python3

o1 = object()
o2 = object()
l  = list()

if o1 not in l: l.append(o1)
if o2 not in l: l.append(o2)

assert(len(l) == 2)
```

with

```py
#!/usr/bin/env python3

d1 = dict()
d2 = dict()
l  = list()

if d1 not in l: l.append(d1)
if d2 not in l: l.append(d2)

assert(len(l) == 2)
```

WHY??  :facepalm:

mtitov:
As I see it now: before provided solution worked fine, since class `rs.Task` was object-origined (wasn't inherited from builtin classes), but it changed with the latest updates: `class DictMixin(dict)` (as an example with `object` as it was before, and as with `dict` - now). In this case we can add the following method to the class `rs.Task`:
```python
def __eq__(self, task):
   return self._id == task._id
```
and no need to change methods `add` and `remove`, but method `get_task` has to be updated (at least with `if t._id == id:`, but my proposal might be faster ;) very subjective point of view)

andre-merzky:
Yeah, that looks good indeed - nice! :-)

andre-merzky:
@mtitov : I commit a version close to your proposal - can you have a look, please?  Thanks!

src/radical/saga/task.py
```diff
@@ -94,6 +94,8 @@ def __init__ (self, _adaptor, _method_type, _method_context, _ttype) :
         self._attributes_set_getter(c.STATE,     self.get_state)
         self._attributes_set_setter(c.STATE,     self._set_state)
 
+        self._uid = ru.generate_id('task.')
```
mtitov, line 97: `generate_id` already creates ID for Task (from inherited `rs.SimpleBase`, [link](https://github.com/radical-cybertools/radical.saga/blob/devel/src/radical/saga/base.py#L47)) -> `self._id`

src/radical/saga/task.py
```diff
@@ -121,6 +123,34 @@ def __init__ (self, _adaptor, _method_type, _method_context, _ttype) :
...
+
+
+    # --------------------------------------------------------------------------
+    #
+    def get_id(self):
+
+        # implement get_id for __eq__
+        return self._uid
+
+
```
mtitov, line 153: could we simplify this? ID is required and it is always set
```python
def __eq__(self, other):
    if isinstance(other, Task):
        return self._id == other._id
    else:
        return super(Task, self).__eq__(other)
```

src/radical/saga/task.py
andre-merzky: Good catch, thanks!

src/radical/saga/task.py
andre-merzky:
```
t = Task()
if (t == None): 
    ...
```
`None` won't have a get_id, and the code would raise on trying to call it.  Or am I missing something?

mtitov: I actually don't see how task-object can be `None`, when it is created it is always the object of `Task` class (`isinstance(t, Task)`). Don't know could it be changed what it returns, but it always will have Task-related methods and attributes
```
>>> import radical.saga as rs
>>> t=rs.Task()
  File "<stdin>", line 1, in <module>
TypeError: __init__() missing 4 required positional arguments: '_adaptor', '_method_type', '_method_context', and '_ttype'
```

andre-merzky: Yes, sure, that was supposed to be pseudo code.  The point is, you can compare a task object against other types, and those may not have a `get_id()` method.  Here is a complete example:
```py
#!/usr/bin/env python3

class Foo(object):

    def __init__(self, uid):

        self.uid = uid

    def __eq__(self, other):

        return self.uid == other.uid

class Bar(object):
    pass


f1a = Foo(1)
f1b = Foo(1)
f2a = Foo(2)
b   = Bar()

assert(f1a == f1b)
assert(f1a != f2a)
assert(f1b != f2a)

assert(f1a)
assert(b)

b == f1a  # raises
```

which results in:
```sh
./t.py 
  File "./t.py", line 29, in <module>
    b == f1a
  File "./t.py", line 11, in __eq__
    return self.uid == other.uid
AttributeError: 'Bar' object has no attribute 'uid'
```

mtitov: I agree when it is compared task-object and non-task-object, but our case is different:
```python
# inside Task.__eq__
    if isinstance(other, Task):
        return self.get_id() == other.get_id()
```
(and if `other` is a non-task-object then call an inherited method `super(Task, self).__eq__(other)`, as you proposed)

andre-merzky: Fair - thanks.

mtitov:
added couple fixes (id-related), it looks good to me now

andre-merzky:
Thanks! :-)

PR #806, opened 2020-08-17
Job description attribute SystemArchitecture

mtitov:
New attribute `SystemArchitecture` will include information about `CPUArchitecture`, `GPUArchitecture` and `SMT`, e.g. `{'cpu': '<cpu_arch>', 'gpu': '<gpu_arch>', 'smt': <smt_number>}`


andre-merzky:
@mturilli : I suspect the default will vary per resource, not per adaptor, which makes this a wee bit more complicated...  One option would be to universally user the default `1` - that's easy to document :-D

mturilli:
I see, maybe then always use the parameter for each resource so that it self-documents in the config file?

src/radical/saga/adaptors/slurm/slurm_job.py
```diff
@@ -606,11 +609,10 @@ def _job_run(self, jd):
                 # gres resources are specified *per node*
                 assert(n_nodes), 'need unique number of cores per node'
 
-                if cpu_arch: gpu_arch = cpu_arch.lower()
-                else       : gpu_arch = 'p100'
-                if gpu_arch == 'k80': count = 4
-                else                : count = 2
-                script += "#SBATCH --gres=gpu:%s:%d\n" % (gpu_arch, count)  # Make sure we take a full GPU node
+                gpu_arch = 'p100' if not gpu_arch else gpu_arch.lower()
```
andre-merzky, line 612: This is correct, but I don't think it makes the code clearer.

src/radical/saga/adaptors/slurm/slurm_job.py
```diff
@@ -606,11 +609,10 @@ def _job_run(self, jd):
...
+                count = 4 if gpu_arch == 'k80' else 2
```
andre-merzky, line 613: same

src/radical/saga/adaptors/slurm/slurm_job.py
mtitov: it is pretty standard and straightforward construction (reduces space and fulfill pylint/flake-requirements), but ok I'll return previous one back

andre-merzky: *sigh* I know, thanks for bearing with me.

PR #808, opened 2020-08-19
Key `options` for JD attribute SystemArchitecture (LSF)

mtitov:


src/radical/saga/adaptors/lsf/lsfjob.py
```diff
@@ -40,9 +40,17 @@
 # FIXME: will be taken from resource config
 RESOURCES = {
     'summit': {'cpn': 42,
-               'gpn': 6},
+               'gpn': 6,
+               'valid_alloc_flags': [
```
andre-merzky, line 44: Maybe `valid_options`?

andre-merzky (approved):
Yeah, that looks good.   Thanks :-)

src/radical/saga/adaptors/lsf/lsfjob.py
mtitov: that also works, but I have a point why did pick that one - since it is inside of LSF adaptor, here we put all values that are related to `alloc_flags` (the actual values provided by user/config are taken from `jd.system_architecture['options']`)

andre-merzky: Yeah, that's fair, works for me.  Thanks.

PR #809, opened 2020-08-26
fix id based __eq__

andre-merzky:


mtitov:
agree with these made changes and have more to confirm:
currently we have `t.get_id() == id` in `Containe.get_task` [here](https://github.com/radical-cybertools/radical.saga/blob/hotfix/task_eq/src/radical/saga/task.py#L595), and originally it was `t.id` (before I updated it to use `_id` which was taken from `Task.get_id`). Thus would `t.get_id` give `t.id` (which will work only for `Job` instances)? or should we make it explicit:
```python
        for t in self.tasks:
            task_id = getattr(t, 'id', t._id)
            if task_id == id:
                return t
```

andre-merzky:
Yeah, I kind of expected that you stumble over that line :-)  I am a bit hesitant to change more code, and we need to get this out fast.  But then again it looks good...  Please do some small tests and commit if you are confident that the multi-pilot case remains functional.  Small test case:

```py
#!/usr/bin/env python3

import time
import radical.pilot as rp


try:
    session = rp.Session()

    pmgr  = rp.PilotManager(session=session)
    pdesc = rp.ComputePilotDescription({
    	'resource': 'local.debug',
    	'runtime': 60,
    	'cores': 1
    })

    pilots = pmgr.submit_pilots([pdesc] * 4)

    while True:

        time.sleep(1)

        states = [p.state for p in pilots]
        print(states)

        if rp.PMGR_ACTIVE in states:

            time.sleep(60)

            states = [p.state for p in pilots]
            print(states)
            assert(len(set(states)) == 1), states
            break

except:
    session.close(download=False)
    raise
```

Thanks!

mtitov:
@andre-merzky sorry, took some time for me to figure out things! your test works fine, and I see that for `Job` instances attribute `id` contains data from `get_id`. Thus just to keep method `get_id` inside `Container.get_task` is enough (works for `Task`-instances and `Task`-based such as `Job`-instances). All good! Thanks!

PR #811, opened 2020-09-22
fix sh timeout for long running sessions

andre-merzky:
This adds an infrequent keepalive-`ping` between service instance and shell_wrapper so that the shell timeout does not kick in as long as the connection is up.

### DIFF
--- a/src/radical/saga/adaptors/shell/shell_job.py
+++ b/src/radical/saga/adaptors/shell/shell_job.py
@@ -8,7 +8,7 @@ __license__   = "MIT"
 
 import re
 import time
-import threading
+import threading as mt
 
 import radical.utils as ru
 
@@ -25,7 +25,12 @@ from . import shell_wrapper
 
 # ------------------------------------------------------------------------------
 #
-class _job_state_monitor(threading.Thread):
+_PING_DELAY  = 60.0
+
+
+# ------------------------------------------------------------------------------
+#
+class _job_state_monitor(mt.Thread):
     ''' thread that periodically monitors job states '''
 
     # --------------------------------------------------------------------------
@@ -491,10 +496,11 @@ class ShellJobService(cpi.Service):
             #   cmd_state() { touch $DIR/purgeable; ... }
             # When should that be done?
 
-         #  self.shell.run_sync("PURGE", iomode=None)
-            self.shell.run_async("QUIT")
-            self.shell.finalize(kill_pty=True)
-            self.shell = None
+            with self._shell_lock:
+             #  self.shell.run_sync("PURGE", iomode=None)
+                self.shell.run_async("QUIT")
+                self.shell.finalize(kill_pty=True)
+                self.shell = None
 
         if self.monitor:
             self.monitor.finalize()
@@ -505,9 +511,27 @@ class ShellJobService(cpi.Service):
     #
     def initialize(self):
 
+        # the underlying shell is not always ready to receive new commands,
+        # e.g.,  when running an async command and still collecting output.
+        # The driving code is usually aware of the shell's state - but we want
+        # to send `PING` requests now and then to ensure the shell does not
+        # time out, and the thread triggering that ping is *not* aware of the
+        # shell state.  To avoid interfering with async calls, we lock any
+        # interaction with the shell
+        self._shell_lock = mt.RLock()
+
+        # at regular intervals, run a ping toward the shell wrapper to avoid
+        # timeouts kicking in
+        # FIXME: configurable frequency
+        self._ping = mt.Timer(_PING_DELAY, self._ping_cb)
+        self._ping.start()
+
+
         # very first step: get the remote environment, and expand the config
         # settings
-        ret, out, err = self.shell.run_sync(' env')
+        with self._shell_lock:
+            ret, out, err = self.shell.run_sync(' env')
+
         if ret != 0:
             raise rse.NoSuccess("env query failed(%s):(%s)(%s)"
                     % (ret, out, err))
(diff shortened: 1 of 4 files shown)


src/radical/saga/adaptors/shell/shell_job.py
```diff
@@ -622,6 +655,19 @@ def initialize(self):
...
+    #
+    def _ping_cb (self) :
+
+        with self._shell_lock:
+
+            _, out, _ = self.shell.run_sync('PING')
+            assert('PONG' in out), out
+
+            self._ping = mt.Timer(_PING_DELAY, self._ping_cb)
+            self._ping.start()
```
mtitov, line 668: these 2 lines are repeated, `self._ping` was set in `initialization` method (L526)

mtitov:
@andre-merzky should [this loop](https://github.com/radical-cybertools/radical.saga/blob/bd4904ff853f03948a9c57bd0db54ca7e455b88d/src/radical/saga/adaptors/shell/shell_job.py#L1094) be wrapped with `with self._shell_lock:` as well? (similar to L723)

src/radical/saga/adaptors/shell/shell_job.py
```diff
@@ -622,6 +655,19 @@ def initialize(self):
         self._logger.debug("got mon prompt(%s)(%s)" % (ret, out.strip()))
 
 
+    # ----------------------------------------------------------------
+    #
+    def _ping_cb (self) :
+
+        with self._shell_lock:
+
+            _, out, _ = self.shell.run_sync('PING')
```
mtitov, line 664: it happens while `self.shell` is not created yet, but this callback is called (from travis):
```
Exception in thread Thread-50:
842Traceback (most recent call last):
843  File "/opt/python/3.6.7/lib/python3.6/threading.py", line 916, in _bootstrap_inner
844    self.run()
845  File "/opt/python/3.6.7/lib/python3.6/threading.py", line 1180, in run
846    self.finished.wait(self.interval)
847  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/radical/saga/adaptors/shell/shell_job.py", line 664, in _ping_cb
848    _, out, _ = self.shell.run_sync('PING')
849AttributeError: 'NoneType' object has no attribute 'run_sync'v
```
just add before this line `if self.shell:`?

andre-merzky: Then the test should call `init_instance` I guess?  SAGA has delayed instantiation for most classes, to support async instance creation.  So, unless `init_instance` is called, the class is not considered viable, and cannot really be tested either.

src/radical/saga/adaptors/shell/shell_job.py
andre-merzky: That's on purpose: the `self_ping` timer can only be called once - so whenever it is called, it re-initializes a new timer instance for the next call.  I don't like it either, but that's how the threading.Timer class works _*shrug*_

src/radical/saga/adaptors/shell/shell_job.py
mtitov: yeah, that also works to move it, agree with that

mtitov:
btw to pass tests I had to clean up `$HOME/.radical/saga` (old wrapper wasn't rewritten)

andre-merzky:
> btw to pass tests I had to clean up `$HOME/.radical/saga` (old wrapper wasn't rewritten)

Hmm, probably not good, but I would like to keep that fix out of this PR...

src/radical/saga/adaptors/shell/shell_job.py
```diff
@@ -396,7 +397,9 @@ def stage_input(self, shell, jd):
                     source = local
                     target = remote
                     self._logger.info("stage %s to %s" % (source, target))
-                    shell.stage_to_remote(source, target)
+
+                    with self._shell_lock:
```
mtitov, line 401: I think it shouldn't be here, `self._shell_lock` is not defined in class `Adaptor`, but in `ShellJobService`

src/radical/saga/adaptors/shell/shell_job.py
```diff
@@ -417,7 +420,9 @@ def stage_output(self, shell, jd):
                     source = remote
                     target = local
                     self._logger.info("stage %s to %s" % (source, target))
-                    shell.stage_from_remote(source, target)
+
+                    with self._shell_lock:
```
mtitov, line 424: same here as above

src/radical/saga/adaptors/shell/shell_job.py
mtitov: I read your comment incorrectly first... to call `init_instance` manually would help for test, but I didn't figure out how to put it there. But to check that `self.shell` doesn't contain None value shouldn't break anything

andre-merzky: > But to check that self.shell doesn't contain None value shouldn't break anything

well, no - but that implies that we would need to sprinkle such tests through the whole class to check for everything which is set up in the initializer, which is something which would obfuscate the code a lot, only to shield for missing initialization (and would then have to fail anyway).  So I'd rather not do that...

src/radical/saga/adaptors/shell/shell_job.py
andre-merzky: Yeah, that was quite stupid - is fixed now, the locks are passed to the calls.

src/radical/saga/adaptors/shell/shell_job.py
mtitov: That errors actually were after closing JobService... since Timer was still on. I've added `self._ping.cancel()` inside `close()` method

mtitov (approved):

src/radical/saga/adaptors/shell/shell_job.py
andre-merzky: Ah, that makes sense.  Thanks!

PR #812, opened 2020-09-29
Fix for ShellJob[Service]

mtitov:
Inside `ShellJob` there are several methods (`get_stdout`, `get_stderr`, `get_log`) that have a call `self.js.shell.run_sync`, should it be put into `with self.js._shell_lock:` construction?


andre-merzky:
Yes, they should, thanks for catching that!  They would otherwise interfere with the `ping` thread (potentially).

Issue #813, opened 2020-10-22
--gres flag on xsede.comet gpu


lee212:
Similarily to #717, SDSC Comet uses `--gres` flag instead of `--gpus` because I see:
```
sbatch: unrecognized option '--gpus=4'
```

PR #814, opened 2020-11-20
comet takes --gres for gpu

lee212:
This is to reflect slurm options for gpus, and addressing issue #813 according to https://portal.xsede.org/sdsc-comet#gpu.

### DIFF
--- a/src/radical/saga/adaptors/slurm/slurm_job.py
+++ b/src/radical/saga/adaptors/slurm/slurm_job.py
@@ -625,11 +625,13 @@ class SLURMJobService(cpi_job.Service):
                 # gres resources are specified *per node*
                 assert(n_nodes), 'need unique number of cores per node'
 
-                if gpu_arch: gpu_arch = gpu_arch.lower()
-                else       : gpu_arch = 'p100'
+                # if no `gpu_arch` then first available gpu node (either type)
+                # gpu types are "p100" and "k80"
+                if gpu_arch: gpu_arch_str = ':%s' % gpu_arch.lower()
+                else       : gpu_arch_str = ''
                 count = 4
                 # Make sure we take a full GPU node
-                script += "#SBATCH --gres=gpu:%s:%d\n" % (gpu_arch, count)
+                script += "#SBATCH --gres=gpu%s:%d\n" % (gpu_arch_str, count)
 
         elif 'tiger' in self.rm.host.lower():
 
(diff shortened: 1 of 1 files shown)


mtitov:
@lee212 please check this [branch](https://github.com/radical-cybertools/radical.saga/tree/fix/comet_gpu), it covers all cases (for `gpu` queue, and not for `gpu-share` queue). It was mentioned [here](https://github.com/radical-cybertools/radical.pilot/issues/2259#issuecomment-730684666)

lee212:
@mtitov , check the code block for gpu counting. On `gpu` queue, you have to increase gpu counts by 4, a number of available gpu devices per node. For example, 1,2, and 3 will be failed on `gpu` queue but okay with `gpu-shared` queue.

mtitov:
@lee212 yeah, just I didn't see it explicitly defined at user guide, but if you checked that then ok. Also don't forget to consider option when `gpu_arch` is not provided

src/radical/saga/adaptors/slurm/slurm_job.py
```diff
@@ -619,6 +619,18 @@ def _job_run(self, jd):
...
+
+            if gpu_count:
+                # gres resources are specified *per node*
+                assert(n_nodes), 'need unique number of cores per node'
+
+                if gpu_arch: gpu_arch = gpu_arch.lower()
+                else       : gpu_arch = 'p100'
+                count = 4
+                # Make sure we take a full GPU node
+                script += "#SBATCH --gres=gpu:%s:%d\n" % (gpu_arch, count)
```
mtitov, line 632: proposing the following changes
```python
                    gres_list = ['gpu']
                    if gpu_arch: gres_list.append(gpu_arch.lower())
                    gres_list.append(str(count))
                    script += "#SBATCH --gres=%s\n" % ':'.join(gres_list)
```
to consider the option when `gpu_arch` is not set

andre-merzky: Wouldn't that change semantics?  At the moment, `p100` is set when `gpu_arch` is unset.  Also, I don't think the changed code is easier to read.

mtitov: > At the moment, p100 is set when gpu_arch is unset

right, but it is possible to let the system pick the gpu, e.g., here what they provide in the documentation

> #SBATCH --gres=gpu:4        #first available gpu node (either type)
> #SBATCH --gres=gpu:k80:4    #only K80 nodes
> #SBATCH --gres=gpu:p100:4   #only P100 nodes

p.s. and yeah, code could be changed

andre-merzky: > let the system pick the gpu

I see, makes sense.   @lee212 : would that work for you?

lee212: Yes, thanks for the changes

andre-merzky (approved):
LGTM, thanks!

PR #815, opened 2020-11-25
ip-isolate flag for distributed pytorch on lassen

lee212:
This PR aims to introduce `"-alloc_flags ipisolate"` for lsf jobs on Lassen. One use case is that distributed pytorch requires sockets opened without authentication, which isn't abiding LC security policy. This flag is accommodating with the system, and for more details: https://lc.llnl.gov/confluence/pages/viewpage.action?pageId=650674651


andre-merzky (approved):
I see a bottomless pit opening up there... :-/  Don't see an alternative either though...

mtitov:
I would think that after switching to Config class, this could be moved into `_verify` method

PR #818, opened 2020-12-26
Fix/issue 787

andre-merzky:
fix creation flags. ensure `None` on missing exit code


src/radical/saga/adaptors/shell/shell_file.py
```diff
@@ -518,7 +518,8 @@ def copy(self, src_in, tgt_in, flags, _from_task=None):
         if  rsumisc.url_is_compatible(cwdurl, src) and \
             rsumisc.url_is_compatible(cwdurl, tgt):
 
-            if  flags & c.CREATE_PARENTS:
+            if  flags & c.CREATE or \
+                flags & c.CREATE_PARENTS:
```
mtitov, line 522: should we set the order explicitly?
either
```python
            if  (flags & c.CREATE) or \
                (flags & c.CREATE_PARENTS):
```
or
```python
            if  flags & (c.CREATE or c.CREATE_PARENTS):
```
\* reference to the [documentation](https://docs.python.org/3/reference/expressions.html#operator-precedence)

andre-merzky: I don't think the second version works - that would create `True` for `(c.CREATE or c.CREATE_PARENTS)` (since neither is `0`), and we would constantly do `flags & True` which happens to always return `1` for uneven `flag` integers. Funny, but not what we want :-)  Maybe `flags & (CREATE | CREATE_PARENTS)`?  We can do the first one to clarify precedence - will change!

src/radical/saga/adaptors/shell/shell_file.py
```diff
@@ -641,7 +644,8 @@ def link(self, src_in, tgt_in, flags, _from_task=None):
         if  flags & c.RECURSIVE:
             raise rse.BadParameter("'RECURSIVE' flag not  supported for link()")
 
-        if  flags & c.CREATE_PARENTS:
+        if  flags & c.CREATE or \
+            flags & c.CREATE_PARENTS:
```
mtitov, line 648: one leftover

andre-merzky: Thanks!

mtitov (approved):
LGTM

PR #822, opened 2021-02-11
Bridges2 support

iparask:
Bridges 2.

PS: Can we reopen the discussion about GPUs and how we define them?

### DIFF
--- a/src/radical/saga/adaptors/slurm/slurm_job.py
+++ b/src/radical/saga/adaptors/slurm/slurm_job.py
@@ -604,7 +604,14 @@ class SLURMJobService(cpi_job.Service):
         # target host specifica
         # FIXME: these should be moved into resource config files
         self._logger.debug ("submit SLURM script to %s", self.rm)
-        if 'bridges' in self.rm.host.lower():
+        if 'bridges2' in self.rm.host.lower():
+             if gpu_count:
+                 # gres resources are specified *per node*
+                 assert(n_nodes), 'need unique number of cores per node'
+
+                 script += "#SBATCH --gres=gpu:v100:8\n"
+
+        elif 'bridges' in self.rm.host.lower():
 
             if gpu_count:
                 # gres resources are specified *per node*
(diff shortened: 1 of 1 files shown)


src/radical/saga/adaptors/slurm/slurm_job.py
```diff
@@ -604,7 +604,14 @@ def _job_run(self, jd):
         # target host specifica
         # FIXME: these should be moved into resource config files
         self._logger.debug ("submit SLURM script to %s", self.rm)
-        if 'bridges' in self.rm.host.lower():
+        if 'bridges2' in self.rm.host.lower():
+             if gpu_count:
+                 # gres resources are specified *per node*
+                 assert(n_nodes), 'need unique number of cores per node'
+
+                 script += "#SBATCH --gres=gpu:v100:8\n"
```
andre-merzky, line 612: Hmm, shouldn't this trickle down from the `system_architecture` setting in the RP config?

andre-merzky (commented):
thanks!

src/radical/saga/adaptors/slurm/slurm_job.py
iparask: Yes. Let me find how to do it and update

iparask: I have a question. RP's config defines `system_architecture` and `gpus_per_node`. RS job description includes the `system_architecture`, but has no means to get the number of GPUs per node. May I suggest the following:

```suggestion
                 script += "#SBATCH --gres=gpu:%s:%s\n" % (system_arch.split(':'))
```

where in the config `system_architecture` is `{"gpu":"v100:8"}.

What do you think?

mtitov: btw there is an internal variable that reads `'gpu'` from `system_architecture` -> `gpu_arch`

iparask:
ready!

mtitov (approved):

andre-merzky (approved):

PR #823, opened 2021-03-05
update slurm adapter for ORNL andes cluster

wjlei1990:


mtitov:
hi @wjlei1990, can you please try to add `'andes' in self.rm.host.lower()` to the line 577 instead of 583 (right after `longhorn`)?

wjlei1990:
For now, I don't set "--ntasks" or "-n" in my slurm header. I don't know if it is needed or not...

mtitov:
RP will set a number of requested cores for this option (as an upper threshold). Don't have a strong opinion about this, since the following option `--ntasks-per-node` will be set (we don't have a resource config for Andes yet, but `cores_per_node` will be set there and that corresponding value will be used for `--ntasks-per-node`).

andre-merzky (approved):
LGTM

PR #827, opened 2021-05-20
Cobalt script option `--attrs`

mtitov:


src/radical/saga/adaptors/cobalt/cobaltjob.py
```diff
@@ -168,46 +156,37 @@ def _cobaltscript_generator(url, logger, jd, ppn, is_cray=False, queue=None,
...
         hours = int(jd.wall_time_limit / 60)
         minutes = jd.wall_time_limit % 60
         cobalt_params += '#COBALT --time %s:%s:00\n' \
             % (str(hours).zfill(2), str(minutes).zfill(2))
 
-    if queue:
-        cobalt_params += '#COBALT --queue %s\n' % queue
-    elif jd.queue:
-        cobalt_params += '#COBALT --queue %s\n' % jd.queue
+    cobalt_params += '#COBALT --queue %s\n' % (queue or jd.queue)
```
andre-merzky, line 189: What happens if neither is set?

src/radical/saga/adaptors/cobalt/cobaltjob.py
```diff
@@ -271,22 +249,21 @@ def _cobaltscript_generator(url, logger, jd, ppn, is_cray=False, queue=None,
...
-
     # Set number of nodes
     cobalt_params += '#COBALT --nodecount %d\n' % number_of_nodes
 
     # Set the total number of processes
     cobalt_params += '#COBALT --proccount %d\n' % number_of_processes
 
-    # The Environments are added at the end because for now
-    # Cobalt isn't supporting spaces in the env variables...
-    # Which mess up the whole script if they are at the begining of the list...
```
andre-merzky, line 284: Is that information outdated?

src/radical/saga/adaptors/cobalt/cobaltjob.py
```diff
@@ -271,22 +249,21 @@ def _cobaltscript_generator(url, logger, jd, ppn, is_cray=False, queue=None,
...
+    cobalt_attrs = jd.system_architecture.get('options', [])
+    if cobalt_attrs:
+        cobalt_params += '#COBALT --attrs %s\n' % ':'.join(cobalt_attrs)
+
```
andre-merzky, line 261: Do you have an example of attributes intended to be passed this way?  Thanks!

src/radical/saga/adaptors/cobalt/cobaltjob.py
mtitov: ah, valid, I misread original statement (as `if...else`), will fix, thanks

src/radical/saga/adaptors/cobalt/cobaltjob.py
mtitov: I didn't find either confirming or opposite, thus thought since we keep such order of parameters then it's ok to eliminate this comment. Should I bring it back?

src/radical/saga/adaptors/cobalt/cobaltjob.py
mtitov: I didn't research it much, but have some that are in test script (`mcdram=cache:numa=quad:mig-mode=True`). 

I didn't add here a list of valid attributes (similar to Summit and Lassen), since here, besides attributes themselves, are values of that attributes, which could be more dynamic

src/radical/saga/adaptors/cobalt/cobaltjob.py
andre-merzky: I don't understand the original comment really.  How would 
```
#COBALT FOO=BAR BUZ
#COBALT -queue foo
```
be a problem?  I can understand that the `env` might be a problem, but I don't see why the order would be.  
So, I am not sure if that is worthwhile keeping *shrug* I am fine either way.

src/radical/saga/adaptors/cobalt/cobaltjob.py
andre-merzky: Thanks

andre-merzky (approved):

PR #831, opened 2021-06-29
simpler workdir switch, cleanup

### LINKED ISSUE #830: working_directory in slurm
The slurm adaptor currently has some hardcoded exceptions for switching between the 4 years old old behavior of `--workdir` and `--chdir`. Its probably time to change the default. What is probably an even better solution is to use the shorthand `-D` as that works for both versions.

andre-merzky:
@marksantcroos: would you mind giving this a try?  Thanks!
This fixes #830


mtitov (approved):
LGTM

PR #841, opened 2022-01-25
consistent use of radical_base

### LINKED ISSUE #838: File permission err for non-default .radical/ path in ptyshell
Hi ! 

I'm working on a web interface wrapping saga but I have configuration problem.
I'm running my app with django on apache. 
Without any conf, my home is /var/www/ , so /var/www/.radical/*  raise permission error
I tried to set RADICAL_BASE and it partially worked, as well as RADICAL_SAGA_LOG_TGT in /var/www/html/<my_project>

But when I tried to initiate a shell connection I got an error about /var/www/.radical/ again.
With further tracing I've ended finding this : 

https://github.com/radical-cybertools/radical.saga/blob/ed6634ff8d7b31ff02b5f23d2aad1189d0f73360/src/radical/saga/utils/pty_shell.py#L228-L230

Is it possible to override this ? Maybe using conf file ?

Besides, is there a way to display all configs values ?

andre-merzky:
This PR removes remnants of `$HOME/.radical` throughout the code base and consistently uses `$RADICAL_BASE/.radical` instead.  Fixes #838

### DIFF
--- a/src/radical/saga/adaptors/shell/shell_job.py
+++ b/src/radical/saga/adaptors/shell/shell_job.py
@@ -263,7 +263,7 @@ _ADAPTOR_DOC           = {
           * number of files are limited, as is disk space: the job.service will
 
             keep job state on the remote disk, in
-            ``~/.radical/saga/adaptors/shell_job/``.
+            `$RADICAL_BASE/.radical/saga/adaptors/shell_job/`.
             Quota limitations may limit the number of files created there,
             and/or the total size of that directory.
 
@@ -557,8 +557,8 @@ class ShellJobService(cpi.Service):
         # expand those config entries we want to use(where needed)
         self.notifications  = cfg.enable_notifications
         self.purge_on_start = cfg.purge_on_star
-        self.base_workdir   = ru.expand_env(cfg.base_workdir, env)
 
+        self.base_workdir   = ru.get_radical_base() + '/saga/adaptors/shell_job'
 
         # start the shell, find its prompt.  If that is up and running, we can
         # bootstrap our wrapper script, and then run jobs etc.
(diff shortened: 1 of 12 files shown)


mtitov (approved):
LGTM!

src/radical/saga/adaptors/shell/shell_job.py
```diff
@@ -568,6 +568,11 @@ def initialize(self):
         # stdio.
 
         base = self.base_workdir
+        with self._shell_lock:
+            ret, out, _ = self.shell.run_sync(' mkdir -p %s' % base)
+
+        if ret != 0:
+            raise rse.NoSuccess('host setup failed (%s):(%s)' % (ret, out))
```
andre-merzky, line 575: shouldn't the `mkdir` covered by `ru.get_radical_base()`?

src/radical/saga/utils/pty_shell.py
```diff
@@ -225,20 +225,12 @@ def __init__ (self, url, session=None, logger=None, cfg=None, posix=True,
...
-
+            ru.rec_makedir(self.base)
         except OSError as e:
-            if e.errno == errno.EEXIST and os.path.isdir (self.base):
-                pass
-            else:
-                raise rse.NoSuccess ("could not create staging dir: %s" % e) \
-                      from e
-
+            raise rse.NoSuccess('could not create staging dir: %s' % e) from e
```
andre-merzky, line 233: Same

andre-merzky:
Thanks for cleaning this up! :-)  Added a comment though...

src/radical/saga/adaptors/shell/shell_job.py
mtitov: that is done for base dirs (inside `ru.get_radical_base()`), not for path added after,
kinda thought that having `get_radical_base('saga.adaptors.shell_job')` could be confusing, but I'm ok if you would want to switch to this format
and another thing, since here is used `self.shell.run_sync` (synced command execution), I wasn't sure should we switch from it (but also ok to switch considering to my comment above)

andre-merzky: Ah, of course - thanks.  Yeah, this LGTM then :-)

andre-merzky:
Merge will wait for feedback on #841 to confirm this does the expected for @CalmingBreeze

PR #847, opened 2022-04-06
Feature/slurm notifications

andre-merzky:
Now all relevant(*) job adaptors have async state notifications and we can simplify RPs pilot launcher.

(*) missing in condor, sge and loadleveler - but I doubt we'll update those ever.

### DIFF
--- a/src/radical/saga/adaptors/slurm/slurm_job.py
+++ b/src/radical/saga/adaptors/slurm/slurm_job.py
@@ -561,7 +561,7 @@ class SLURMJobService(cpi_job.Service):
         runs a job on the wrapper via pty, and returns the job id
         '''
 
-        jd = job_obj.jd
+        jd = job_obj.get_description()
 
         # define a bunch of default args
         exe                 = jd.executable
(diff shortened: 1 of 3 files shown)


src/radical/saga/adaptors/slurm/slurm_job.py
```diff
@@ -296,6 +375,9 @@ def init_instance(self, adaptor_state, rm_url, session):
         self.rm      = rm_url
         self.session = session
 
+        self.mt = _job_state_monitor(job_service=self)
+        self.mt.start()
```
mtitov, line 379: it never stops here

src/radical/saga/adaptors/slurm/slurm_job.py
```diff
@@ -674,7 +674,6 @@ def _job_run(self, job_obj):
 
                 assert(n_nodes), 'need unique number of cores per node'
                 script += "#SBATCH -N %d\n" % n_nodes
-                script += "#SBATCH -n %d\n" % n_procs
```
mtitov: I've revert it back, since this section is with option `-N` only, but previous section has both options `-N` and `-n`

src/radical/saga/adaptors/slurm/slurm_job.py
andre-merzky: That's ok - the thread is configured as daemon thread and will die with the process.

mtitov (approved):
LGTM

PR #848, opened 2022-04-13
Fix/slurm spock

andre-merzky:
Should this be merged?


src/radical/saga/adaptors/slurm/slurm_job.py
```diff
@@ -663,6 +662,8 @@ def _job_run(self, job_obj):
 
             if  'stampede2' in self.rm.host.lower() or \
                 'longhorn'  in self.rm.host.lower() or \
+                'spock'     in self.rm.host.lower() or \
```
mtitov, line 665: should we move it into the section below with just `-N` option? (since examples with Spock are not with `-n`)

andre-merzky: Sure - that's done now.

PR #851, opened 2022-09-07
Enforce to use a particular directory while copying batch script to the remote machine

mtitov:


src/radical/saga/adaptors/slurm/slurm_job.py
```diff
@@ -814,10 +814,12 @@ def _job_run(self, job_obj):
         self._logger.info("SLURM script generated:\n%s" % script)
 
         tgt = os.path.basename(tempfile.mktemp(suffix='.slurm', prefix='tmp_'))
+        if cwd:
+            tgt = os.path.join(cwd, tgt)
         self.shell.write_to_remote(src=script, tgt=tgt)
 
         # submit the job
-        ret, out, _ = self.shell.run_sync("sbatch '%s'; echo rm -f '%s'" % (tgt,tgt))
+        ret, out, _ = self.shell.run_sync('sbatch %s && rm -f %s' % (tgt, tgt))
```
andre-merzky, line 822: Putting the file into the job's work dir makes sense I guess - but then we don't really need t remove it anymore, do we?  Anyway, this LGTM.

andre-merzky (approved):

src/radical/saga/adaptors/slurm/slurm_job.py
mtitov: > but then we don't really need t remove it anymore, do we?

agree, we can keep it. Btw if to keep it, then would it be ok to remove line L814 with logging its content? (since we already will have file, and its path will be in log)

andre-merzky: Yes, makes sense.

PR #854, opened 2022-09-12
keep in sync with RADICAL-Pilot SMT setup

mtitov:


src/radical/saga/adaptors/lsf/lsfjob.py
```diff
@@ -234,6 +234,8 @@ def _lsfscript_generator(url, logger, jd, ppn, lsf_version, queue):
     if smt not in SMT_VALID_VALUES:
         smt = SMT_DEFAULT
 
+    assert (int(jd.environment.get('RADICAL_SMT', 1)) == smt)
```
andre-merzky, line 237: I am not sure about the assert.  For example, `smt` might not be set in the system arch.  Should we really assert if `RADICAL_SMT` is different than the default value in this case - shouldn't we just use the env value?

mtitov: Didn't want to have them out of sync between each other. But I agree to have env variable with higher priority, and then if it is not set, then check SMT from `system_architecture` (and remove assert)

p.s. and it would be truly in sync with RP, since RP checks env variable first

committed accordingly

src/radical/saga/adaptors/lsf/lsfjob.py
```diff
@@ -252,8 +255,8 @@ def _lsfscript_generator(url, logger, jd, ppn, lsf_version, queue):
     alloc_flags.append('smt%d' % smt)
     lsf_bsubs += "#BSUB -alloc_flags '%s' \n" % ' '.join(alloc_flags)
 
-    env_string += "export RADICAL_SAGA_SMT=%d" % smt
     if jd.environment:
+        env_string += "export"
```
andre-merzky, line 259: So we don't export `RADICAL_SAGA_SMT` anymore, which is fine IMHO - but should't we export `RADICAL_SMT` instead?  Or maybe we do somewhere else in the code already?  The change to the test seems to indicate no, as the env is explicitly added to the job description?

andre-merzky (approved):
Left one more comment, but otherwise LGTM - thanks!

src/radical/saga/adaptors/lsf/lsfjob.py
mtitov: > So we don't export `RADICAL_SAGA_SMT` anymore

right, at some point it became a leftover

> but should't we export `RADICAL_SMT` instead?

this will be controlled by RP, so RP provides this env variable within `jd.environment` and then it is added to this export-statement

andre-merzky: > > but should't we export `RADICAL_SMT` instead?
> 
> this will be controlled by RP, so RP provides this env variable within `jd.environment` and then it is added to this export-statement

True - but since the SAGA layer knows about it, we could leave it to SAGA to export that env in general.  Well, that's out of scope for this PR I guess, just saying...

PR #856, opened 2022-10-04
allow to trap `SIGTERM` before `SIGKILL` is sent

andre-merzky:
trapping is too slow it seems, and we need to delay juuuuust a bit before sending `SIGKILL`.  POSIX has no sub-second sleep, so we (ab)use `/dev/urandom` for a little busy-time.

This supports radical-cybertools/radical.pilot/pull/2725


AymenFJA (approved):
LGTM.

PR #857, opened 2022-11-05
SAGA.Exception should not overwrite representation

andre-merzky:
The SAGA exception overwriting the `repr` and `str` methods makes it impossible to *programmatically* inspect the raised exception.  They are also against Python conventions, so this PR removes them - SAGA now (correctly) relies on the `Exception` base class to do the right thing...


mtitov (approved):
LGTM

PR #863, opened 2022-12-14
enhance SMT control in Slurm adaptor 

mtitov:
Same approach as for LSF adaptor - check env variable `RADICAL_SMT` first


andre-merzky (approved):
LGTM, thanks!

PR #873, opened 2023-03-21
add some job attributes to the shell adaptor

andre-merzky:
but ignore them.  This makes job descriptions easier to port between batch backends and the shell backend.


mtitov (approved):
LGTM! Thanks!

PR #878, opened 2023-05-31
fixed option(s) for PBSPro tool(s)

mtitov:


mtitov:
> I know there aren't many PBSPro hosts left - but are we sure we don't have any where the `-f1` flag would still be needed?

yeah! `-fx` covers all cases (I actually didn't find any example with `-f1`)

src/radical/saga/adaptors/pbspro/pbsprojob.py
```diff
@@ -838,8 +835,6 @@ def _job_get_info(self, job_id, reconnect):
             if job_info.get('exec_host'):
                 job_info['exec_hosts'] = job_info['exec_hosts'].split('+')
 
-            # TORQUE doesn't allow us to distinguish DONE/FAILED on final state
-            # alone, we need to consider the exit_status.
```
andre-merzky, line 842: Not sure why the comment was removed?

src/radical/saga/adaptors/pbspro/pbsprojob.py
```diff
@@ -852,13 +847,6 @@ def _job_get_info(self, job_id, reconnect):
                 and not job_info['end_time']:
                 job_info['end_time'] = time.time()
 
-
-        # PBSPRO state does not indicate error or success -- we derive that from
-        # the exit code
-        if job_info['returncode'] not in [None, 0]:
-            job_info['state'] = api.FAILED
-
-
```
andre-merzky, line 861: Are we sure we don't need this anymore?

andre-merzky (approved):
LGTM overall, but left two questions.  Thanks!

src/radical/saga/adaptors/pbspro/pbsprojob.py
mtitov: some cleanup, since it's for Torque, but this is PBSPro

src/radical/saga/adaptors/pbspro/pbsprojob.py
mtitov: yeah, we have the following place, which should be used to handle the state

https://github.com/radical-cybertools/radical.saga/blob/4088cb1d27841a1600a7d7854c0de23f2680eef6/src/radical/saga/adaptors/pbspro/pbsprojob.py#L839

and this would just overwrite the determined status

src/radical/saga/adaptors/pbspro/pbsprojob.py
andre-merzky: Ah, then never mind, it's a leftover then indeed...

PR #881, opened 2023-07-05
Provide a target for Sphinx :py:mod: role.

eirrgang:
Establish a docutils target for `radical.saga`. Allows the root package module to be linked from inside or outside of the project (via intersphinx through the `objects.inv` file).

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -1,3 +1,5 @@
+.. py:module:: radical.saga
+
 ####################################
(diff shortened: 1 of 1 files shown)


eirrgang:
There is no module docstring in `src/radical/saga/__init__.py`, so the top of the main page seems like a reasonable target for links to `radical.saga`.

andre-merzky (approved):

